PR #5, opened 2016-12-30
fix: read_bytes

adamiwaniuk:
on MSVC &result[len] is illegal and cause crash (assert)


GreyCat:
Thanks!

Confirming that all tests seem to work, and, according 4 benchmarks that we have so far show no visible effect on performance.

PR #7, opened 2017-02-07
kaitai::kstream::reverse for implementing reverse in cpp for strings.

tschoening:
This is related to the following pull request:

https://github.com/kaitai-io/kaitai_struct_compiler/pull/69#issuecomment-278107742

I don't have a working C++ dev environment to be able to compile and/or test this and no time to create one. So it would be great if someone else could test/finish this. Else simply ignore it...

### DIFF
--- a/kaitai/kaitaistream.cpp
+++ b/kaitai/kaitaistream.cpp
@@ -501,6 +501,13 @@ std::string kaitai::kstream::to_string(int val) {
     return ret;
 }
 
+#include <algorithm>
+std::string kaitai::kstream::reverse(std::string val) {
+    std::reverse(val.begin(), val.end());
+
+    return val;
+}
+
 // ========================================================================
 // Other internal methods
 // ========================================================================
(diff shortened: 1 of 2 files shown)


GreyCat:
As far as I see, you're applying `std::reverse` to the underlying backing array of `std::string`, which is probably not a good thing to do — it will modify original string:

```yaml
seq:
  - id: orig_str
    type: strz
    encoding: UTF-8
instances:
  mod_str:
    value: orig_str.reverse
```

I.e. after first call to `mod_str`, `orig_str` will be reversed as well.

[This SO answer](http://stackoverflow.com/a/4952085/487064) says that string should be explicity copied first as well.

tschoening:
> As far as I see, you're applying std::reverse to the underlying backing array of std::string, which is probably not a good thing to do — it will modify original string:

Doesn't the call to `kstream::reverse` itself copies the argument already? I don't use a pointer or reference for exactly that reason. So the function should get a private copy of the argument, reverses it in-place and returns a copy again, depending on compiler optimizations and such.

> 2) Modifying the string but not wanting the caller to see that change. Passing it in by value is preferable: (std::string)

http://stackoverflow.com/a/10789797/2055163

GreyCat:
Ah, my bad, I've overlooked that. Just made the test and it works. Thanks!

PR #9, opened 2017-02-26
correct reading bitfields with sizes between 32bit and 64bit

modernikarus:
The line
uint64_t mask = (1 << n) -1;
might lead to unexpected results. "1" might be of type (signed) integer. An integer might have 32 bits or less. (1 << n) will have the same datatype as "1". This causes unintended behavior for n >= 32.

let type that is left-shiftet to generate a bitmask be of type uint64_t so that bitfields with length 32 to 63 are read properly; add special case for 64 bitfield case; more than 64 bits are not supported at the moment

### DIFF
--- a/kaitai/kaitaistream.cpp
+++ b/kaitai/kaitaistream.cpp
@@ -296,7 +296,8 @@ uint64_t kaitai::kstream::read_bits_int(int n) {
     }
 
     // raw mask with required number of 1s, starting from lowest bit
-    uint64_t mask = (1 << n) - 1;
+    uint64_t mask = ((uint64_t)1 << n) - 1;
+    if (n==64) {mask = 0xFFFFFFFFFFFFFFFF;}
     // shift mask to align with highest bits available in @bits
     int shift_bits = m_bits_left - n;
     mask <<= shift_bits;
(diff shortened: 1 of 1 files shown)


GreyCat:
If the main problem is with that `1` is `signed int`, why don't we use `1UL` then? Let me check if it will solve the problem.

GreyCat:
Thanks for you contribution and detailed analysis of the bug!

Just for sake of documentation, I'll document what I've tried:

* Using `1UL` or `1ULL` is generally not a very good idea:
  * They're not guaranteed to be available in older compilers
  * They represent either "unsigned long" or "unsigned long long", which is not guaranteed to be the same as `uint64_t`
* Indeed, 64-bit masks (i.e. `0xffffffffffffffff`) needs to be treated as special case. All compilers I've tried (gcc, clang, icc, MSVC) treat shift by 64 bits is undefined case (and warn about that, if shift is by a constant) and thus results differ.

@modernikarus Please try to follow general project's code style in your patches :) Also, I've refactored your submission a little to extract a common repeated functionality into a function: see commit c84a867e7107ff2858a5eb9977495ce9240f5d3f.

PR #15, opened 2017-08-21
Link libz and libiconv on Apple platforms

FilippoBiga:


GreyCat:
I believe this patch won't help much at least with the tests, as tests build their own binary with their own CMakeLists. Anyway, final linking decisions would be up to the app, not library, I guess...

FilippoBiga:
This patch allows cloning the repository and building the shared dynamic library `libkaitai_struct_cpp_stl_runtime.dylib` without any additional modification, on macOS. I have no clue how the CI tests are run/handled so I guess it is up to you :)

GreyCat:
Tests are generally located in the tests repo, namely CMakeLists are at https://github.com/kaitai-io/kaitai_struct_tests/blob/master/spec/cpp_stl/CMakeLists.txt - if you could take a look, that would be great :)

I agree that applying this patch is still better than not doing that.

GreyCat:
I've added iconv to tests with APPLE conditional as well, now it fails with that:

```
[100%] Linking CXX executable ks_tests
[100%] Built target ks_tests
Test setup error: boost::runtime::access_to_missing_argument: There is no argument provided for parameter color_output
```

(Details are available at https://travis-ci.org/kaitai-io/ci_targets/jobs/266592259)

PR #17, opened 2017-09-15
Code improvment

crackedmind:
Summary:
1. Fix warnings
2. Pass string as reference
3. Fix crashing when we try to read this structure if readed string size == 0

```yaml
  string:
    seq:
      - id: size
        type: u2
      - id: desc
        type: str
        size: size
```

### DIFF
--- a/kaitai/kaitaistream.cpp
+++ b/kaitai/kaitaistream.cpp
@@ -421,22 +421,22 @@ std::string kaitai::kstream::bytes_terminate(std::string src, char term, bool in
 // ========================================================================
 
 std::string kaitai::kstream::process_xor_one(std::string data, uint8_t key) {
-    int len = data.length();
+    size_t len = data.length();
     std::string result(len, ' ');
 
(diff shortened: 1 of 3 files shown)


GreyCat:
There is a huge controversy / religious question about all that `std::string` vs `const std::string&` passing. As far as I understand, in most modern C++ implementations, there's no copying of variable length contents in any case, the efficiency of moving/switching of pointers is highly debatable.

I'd love to accept everything else, but probably that `const std::string&` switch would require extra discussion, so it's better be done as separate PR, and it would be really helpful to provide more info on that: for example, benchmarks that clearly show benefits of such a switch would be awesome :)

If you don't want to do two separate PRs, I can cherry-pick the rest of your fixes, but it would take some time.

crackedmind:
Ok, reverted this commit for now. I will create separate PR for this.

GreyCat:
Ok, I guess I can just apply them all squashed :)

PR #18, opened 2017-09-18
Added CMake option

garlix:
In this way someone can decide during compile time if use string encoding or not.
To disable string encoding via iconv simply compile with

`build$ cmake -DUSE_STRING_ENCODING=OFF ..`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,8 @@ set (SOURCES
     kaitai/kaitaistream.cpp
 )
 
+option(USE_STRING_ENCODING "Decode strings via iconv" ON)
+
 add_library (${PROJECT_NAME} SHARED ${HEADERS} ${SOURCES})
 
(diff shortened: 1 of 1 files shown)


GreyCat:
Technically, it is supposed to be an option to choose particular string encoding handling algorithm/library. Can we make it a two alternatives choice now (i.e. `OFF` and `ICONV`) and then, when, for example, native Windows API encoding handling would come, we'll add more options (like `WINAPI` or something like that)?

garlix:
Yes, sure! I've only want to make this modification as pull request because I found it useful for myself.
I have updated the `CMakeLists.txt` file so now everybody can compile with

`build$ cmake -DSTRING_ENCODING_TYPE=(ICONV|NONE|...) ..`

default is `ICONV`

GreyCat:
Super! Thanks!

PR #20, opened 2018-04-09
missing throws

arekbulski:
There were few "throw exception" lines missing. Others were just updated to be same form.

### DIFF
--- a/kaitai/kaitaistream.cpp
+++ b/kaitai/kaitaistream.cpp
@@ -344,6 +344,12 @@ uint64_t kaitai::kstream::get_mask_ones(int n) {
 std::string kaitai::kstream::read_bytes(std::streamsize len) {
     std::vector<char> result(len);
 
+    // NOTE: streamsize type is signed, negative values are only *supposed* to not be used.
+    // http://en.cppreference.com/w/cpp/io/streamsize
+    if (len < 0) {
+        throw std::runtime_error("read_bytes: requested a negative amount");
+    }
+
     if (len > 0 ) {
         m_io->read(&result[0], len);
     }
@@ -374,7 +380,7 @@ std::string kaitai::kstream::read_bytes_term(char term, bool include, bool consu
     if (m_io->eof()) {
         // encountered EOF
         if (eos_error) {
-            // throw exception here
+            throw std::runtime_error("read_bytes_term: encountered EOF");
         }
     } else {
         // encountered terminator
@@ -390,9 +396,8 @@ std::string kaitai::kstream::ensure_fixed_contents(std::string expected) {
     std::string actual = read_bytes(expected.length());
 
     if (actual != expected) {
-        std::cout << "Fixed contents mismatch!\n";
-        std::cout << actual << "\n";
-        std::cout << expected << "\n";
+        // NOTE: I think printing it outright is not best idea, it could contain non-ascii charactes like backspace and beeps and whatnot. It would be better to print hexlified version, and also to redirect it to stderr.
+        throw std::runtime_error("ensure_fixed_contents: actual data does not match expected data");
     }
 
     return actual;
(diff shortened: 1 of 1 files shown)


kaitai/kaitaistream.cpp
```diff
@@ -344,6 +344,12 @@ uint64_t kaitai::kstream::get_mask_ones(int n) {
 std::string kaitai::kstream::read_bytes(std::streamsize len) {
     std::vector<char> result(len);
 
+    // NOTE: streamsize type is signed, negative values are only *supposed* to not be used.
+    // http://en.cppreference.com/w/cpp/io/streamsize
+    if (len < 0) {
+        throw std::runtime_error("read_bytes: requested a negative amount");
+    }
```
GreyCat: `read()` would throw a `std::ios_base::failure` exception on negative amount requested, no need to check here.

kaitai/kaitaistream.cpp
```diff
@@ -344,6 +344,12 @@ uint64_t kaitai::kstream::get_mask_ones(int n) {
...
+
     if (len > 0 ) {
```
GreyCat: This check should be actually `len != 0`. Thus we'll do a `read()` with negative amount of bytes, and that will result in clear native C++ exception.

GreyCat: Um, actually, I'm kind of puzzled at why this check is even needed in the first place. At least where I've checked, it looks like it's perfectly ok to call `read()` requesting 0 bytes to read. @crackedmind, could you elaborate?

kaitai/kaitaistream.cpp
```diff
@@ -390,9 +396,8 @@ std::string kaitai::kstream::ensure_fixed_contents(std::string expected) {
     std::string actual = read_bytes(expected.length());
 
     if (actual != expected) {
-        std::cout << "Fixed contents mismatch!\n";
-        std::cout << actual << "\n";
-        std::cout << expected << "\n";
+        // NOTE: I think printing it outright is not best idea, it could contain non-ascii charactes like backspace and beeps and whatnot. It would be better to print hexlified version, and also to redirect it to stderr.
+        throw std::runtime_error("ensure_fixed_contents: actual data does not match expected data");
```
GreyCat: Typically, this is solved by defining a custom exception type and passing actual vs expected to it. It will lay quietly there, and only where end-user would request something like `caught_exception.what()`, we'll render it into a message with hex representations.

For now, though, I guess, throwing a `std::runtime_error` is still much better than that cout placeholder.

kaitai/kaitaistream.cpp
```diff
@@ -565,9 +570,9 @@ std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc)
 
     if (cd == (iconv_t) -1) {
         if (errno == EINVAL) {
-            throw std::runtime_error("invalid encoding pair conversion requested");
+            throw std::runtime_error("bytes_to_str: invalid encoding pair conversion requested");
```
GreyCat: While we've here, actually, this should be `std::invalid_argument`, I guess, as it's clearly caused by invalid encoding being requested.

GreyCat:
Thanks, I've merged it in. Few things I've noted is relatively minor stuff that could be changed later.

kaitai/kaitaistream.cpp
arekbulski: Okie, do you want me to remove this one then?

GreyCat: Don't worry, I'll probably fix all `len`-related checks myself, after the comments requested [here](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/20#discussion_r180395392).

webbnh: Notice that, even before this check, `len` is used to set the initial the size of the vector...I don't know how it will react to a negative value....

GreyCat: Um, we just definitely need more tests for that one, and that question actually still stands for all other target languages.

PR #25, opened 2019-01-25
Add support for write (serialization)

JloveU:
Add support for write (serialization)

### DIFF
--- a/kaitai/kaitaistream.cpp
+++ b/kaitai/kaitaistream.cpp
@@ -27,23 +27,19 @@
 #include <stdexcept>
 
 kaitai::kstream::kstream(std::iostream* io): m_io(io) {
-    init();
+    if (m_io) {
+        init();
+    }
 }
 
-kaitai::kstream::kstream(std::string& data): m_io_str(data) {
-    m_io = &m_io_str;
-    init();
+kaitai::kstream::kstream(const kaitai::kstream &other): m_io(other.m_io), m_bits_left(other.m_bits_left), m_bits(other.m_bits) {
 }
 
 void kaitai::kstream::set_stream(std::iostream* io) {
     m_io = io;
-    init();
-}
-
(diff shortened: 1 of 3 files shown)


GreyCat:
Hi @JloveU! Sorry for taking so long to respond.

Could you elaborate a little on the context of this work? There is a long-running goal to add serialization suport (i.e. infamous https://github.com/kaitai-io/kaitai_struct/issues/27), but it's currently in early design / PoC stages.

I see that you're adding some methods here: some of them are trivial and understandable, some are not. Let us take a step back and talk about implementation plan first in issue 27?

Namely, I'd like to know your opinion on:

1. How to structure API for "write mode" at all — [for Java](https://github.com/kaitai-io/kaitai_struct_java_runtime/tree/serialization/src/main/java/io/kaitai/struct), we've so far opted for separate `KaitaiStruct.ReadOnly` (with `_read()` only) and `KaitaiStruct.ReadWrite` (with `_read()` + `_write()` + `_check()`) abstract classes. 

2. How to structure KaitaiStream API: Java PoC used [separate region for all write-related operations](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/serialization/src/main/java/io/kaitai/struct/KaitaiStream.java#L455).

3. How to make stuff relatively backwards compatible with currently generated API?

JloveU:
I really appreciate the great work of kaitai struct. I'm using it (cpp_stl) for parsing and serializing data which is transferred in network. But unfortunately, there is no serialization available up to now. I have not used Scala before, so I can't modify the kaitai struct compiler. My solution is to add some "write" methods in kstream in cpp_stl runtime, and add "_write()" method manually.

GreyCat:
My point here is that we'll have to solve this for all languages, so we'll need to agree of the concepts first.

For C++, it is actually even more complicated, as we have to consider ownership model, which changed dramatically due to addition of writing. Let us start with 3 questions I've outlined above?

PR #27, opened 2019-06-18
Edit for Mac: runtime/cpp_stl/CMakeLists.txt

idoimaging:
Hi, 

On the Mac I was not able to link the symbols `iconv`, `iconv_open` and `iconv_close` called in `runtime/cpp_stl/kaitaistream.cpp`.  This was due to the `find_library ICONV_LIBRARIES` line in `CMakeLists.txt` looking for `libiconv`, when on the Mac it needs to look for `iconv`.  The same line looks for `c` so it was finding `/usr/lib/libc` and not `/usr/lib/libiconv`, which is installed.  `libc` of course can not resolve the `iconv` symbols.  I took out `c` and added `iconv`.  
I also added two lines to install the headers in a `kaitai` subdirectory (as the generated C++ files look for `kaitai/kaitaistruct.h`).  One line to add `PUBLIC_HEADER` to `set_target_properties`, and one line in the `install` target to install the two `.h` files.
Tested on Mac and Linux but not Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ find_package(ZLIB)
 
 find_path(ICONV_INCLUDE_DIRS iconv.h)
 mark_as_advanced(ICONV_INCLUDE_DIRS)
-find_library(ICONV_LIBRARIES NAMES libiconv libiconv-2 c)
+find_library(ICONV_LIBRARIES NAMES libiconv libiconv-2 iconv)
 mark_as_advanced(ICONV_LIBRARIES)
 
 set(ICONV_FOUND FALSE)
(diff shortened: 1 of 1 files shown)


GreyCat:
I'm merging this in, but please note that this CMakeLists.txt is effectively unmaintained and not really supported. See also https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/22 for gory details.

idoimaging:
Got it, thanks.  I should have read PR 22 earlier.  I made only a few changes and I hope it will help someone else with C++/libiconv issues especially on the Mac.

PR #30, opened 2019-09-07
Initial support for serialization

jchv:
Implements write support for some of the primitives, at least enough to write basic seqs. The API had to be shifted over a bit, but it _should_ be API compatible, though this certainly breaks ABI compatibility. Does this matter? Is there any kind of major version bump to do? I do not think there is a practical way to avoid the ABI compatibility issue.

Currently the code here seems to build just fine, but I'm putting this PR in draft mode since I desire to do more testing and make sure it compiles with existing generated code.


jchv:
Because unit tests hardcode the paths to the cpp files in the runtime, it is necessary to patch the test CMakeLists files.

```
diff --git a/spec/cpp_stl_11/CMakeLists.txt b/spec/cpp_stl_11/CMakeLists.txt
index 1176dab..8b411e6 100644
--- a/spec/cpp_stl_11/CMakeLists.txt
+++ b/spec/cpp_stl_11/CMakeLists.txt
@@ -26,7 +26,9 @@ set(PREREQ_SOURCES
 )
 
 set(RUNTIME_SOURCES
-       ${RUNTIME_SRC_PATH}/kaitai/kaitaistream.cpp
+       ${RUNTIME_SRC_PATH}/kaitai/kio.cpp
+       ${RUNTIME_SRC_PATH}/kaitai/kistream.cpp
+       ${RUNTIME_SRC_PATH}/kaitai/kostream.cpp
 )
 
 add_executable (ks_tests
diff --git a/spec/cpp_stl_98/CMakeLists.txt b/spec/cpp_stl_98/CMakeLists.txt
index f0ae8e8..069f867 100644
--- a/spec/cpp_stl_98/CMakeLists.txt
+++ b/spec/cpp_stl_98/CMakeLists.txt
@@ -22,7 +22,9 @@ set(PREREQ_SOURCES
 )
 
 set(RUNTIME_SOURCES
-       ${RUNTIME_SRC_PATH}/kaitai/kaitaistream.cpp
+       ${RUNTIME_SRC_PATH}/kaitai/kio.cpp
+       ${RUNTIME_SRC_PATH}/kaitai/kistream.cpp
+       ${RUNTIME_SRC_PATH}/kaitai/kostream.cpp
 )
 
 add_executable (ks_tests
```

For me locally, there are some tests that do not compile and one test that fails at master. However, that happens exactly the same regardless of whether I am using this version of the runtime or not, and are evidently fairly unrelated errors, so this PR is probably not broken.

kaitai/kio.cpp
```diff
@@ -0,0 +1,216 @@
+#include <kaitai/kio.h>
+
+#include <iostream>
+#include <stdexcept>
+
```
GreyCat: Probably warrants `using namespace kaitai;` rather than repeating it every time?

jchv: Oh, I was just following the existing convention. What I'd really prefer is just wrapping the file in `namespace kaitai {}` but this doesn't work because of all of the local inclusions. But this works, done.

JPenuchot:
Hello,

Any hope to see these merged in Kaitai someday ? This would pave the way for fixing an old issue in Mixxx [(here)](https://bugs.launchpad.net/mixxx/+bug/1915639).

Regards,
Jules

jchv:
@JPenuchot: For what it's worth, this would only implement runtime primitives necessary for adding basic serialization support into Kaitai Struct for C++. Merging this would be a start, but we would also need to add support into the Kaitai Struct code generation portion as well.

eoan-ermine:
@jchv @GreyCat 

So, what's with this pull request? I saw that the code was reviewed, but it was not merged after that. Is there some kind of blocker?

PR #33, opened 2019-12-17
Fix GCC type warnings

maks500:
Fix sign conversion, sign comparison, strict aliasing and type punning warnings when compiling using GCC.
This should not affect the behavior in any way, kaitai tests results did not change.
Any feedback is welcomed and appreciated.

This pull depends on https://github.com/kaitai-io/kaitai_struct_compiler/pull/187, both of them should be merged or none will work.

### DIFF
--- a/kaitai/kaitaistream.cpp
+++ b/kaitai/kaitaistream.cpp
@@ -77,11 +77,11 @@ bool kaitai::kstream::is_eof() const {
     }
 }
 
-void kaitai::kstream::seek(uint64_t pos) {
+void kaitai::kstream::seek(std::streampos pos) {
     m_io->seekg(pos);
 }
 
-uint64_t kaitai::kstream::pos() {
+std::streampos kaitai::kstream::pos() {
     return m_io->tellg();
 }
 
@@ -90,7 +90,7 @@ uint64_t kaitai::kstream::size() {
     m_io->seekg(0, std::ios::end);
     std::iostream::pos_type len = m_io->tellg();
     m_io->seekg(cur_pos);
-    return len;
+    return kaitai::kstream::pos_to_uint(len);
 }
 
 // ========================================================================
@@ -112,30 +112,30 @@ int8_t kaitai::kstream::read_s1() {
 // ........................................................................
 
 int16_t kaitai::kstream::read_s2be() {
-    int16_t t;
+    uint16_t t;
     m_io->read(reinterpret_cast<char *>(&t), 2);
 #if __BYTE_ORDER == __LITTLE_ENDIAN
     t = bswap_16(t);
 #endif
-    return t;
+    return int16_t(t);
 }
 
 int32_t kaitai::kstream::read_s4be() {
-    int32_t t;
+    uint32_t t;
     m_io->read(reinterpret_cast<char *>(&t), 4);
 #if __BYTE_ORDER == __LITTLE_ENDIAN
     t = bswap_32(t);
 #endif
-    return t;
+    return int32_t(t);
 }
 
 int64_t kaitai::kstream::read_s8be() {
-    int64_t t;
+    uint64_t t;
     m_io->read(reinterpret_cast<char *>(&t), 8);
 #if __BYTE_ORDER == __LITTLE_ENDIAN
     t = bswap_64(t);
 #endif
-    return t;
+    return int64_t(t);
 }
 
 // ........................................................................
@@ -176,7 +176,7 @@ int64_t kaitai::kstream::read_s8le() {
 uint8_t kaitai::kstream::read_u1() {
     char t;
     m_io->get(t);
-    return t;
+    return uint8_t(t);
 }
 
 // ........................................................................
@@ -310,7 +310,7 @@ uint64_t kaitai::kstream::read_bits_int(int n) {
         char buf[8];
         m_io->read(buf, bytes_needed);
         for (int i = 0; i < bytes_needed; i++) {
-            uint8_t b = buf[i];
+            uint8_t b = uint8_t(buf[i]);
             m_bits <<= 8;
             m_bits |= b;
             m_bits_left += 8;
@@ -345,14 +345,13 @@ uint64_t kaitai::kstream::get_mask_ones(int n) {
 // ========================================================================
 
 std::string kaitai::kstream::read_bytes(std::streamsize len) {
-    std::vector<char> result(len);
-
     // NOTE: streamsize type is signed, negative values are only *supposed* to not be used.
     // http://en.cppreference.com/w/cpp/io/streamsize
     if (len < 0) {
         throw std::runtime_error("read_bytes: requested a negative amount");
     }
 
+    std::vector<char> result(static_cast<std::vector<char>::size_type>(len));
     if (len > 0) {
         m_io->read(&result[0], len);
     }
@@ -364,13 +363,13 @@ std::string kaitai::kstream::read_bytes_full() {
     std::iostream::pos_type p1 = m_io->tellg();
     m_io->seekg(0, std::ios::end);
     std::iostream::pos_type p2 = m_io->tellg();
-    size_t len = p2 - p1;
+    std::streamoff len = p2 - p1;
 
     // Note: this requires a std::string to be backed with a
     // contiguous buffer. Officially, it's a only requirement since
     // C++11 (C++98 and C++03 didn't have this requirement), but all
     // major implementations had contiguous buffers anyway.
-    std::string result(len, ' ');
+    std::string result(static_cast<std::string::size_type>(len), ' ');
     m_io->seekg(p1);
     m_io->read(&result[0], len);
 
@@ -396,7 +395,12 @@ std::string kaitai::kstream::read_bytes_term(char term, bool include, bool consu
 }
 
 std::string kaitai::kstream::ensure_fixed_contents(std::string expected) {
-    std::string actual = read_bytes(expected.length());
+    size_t len = expected.length();
+    if (len > std::numeric_limits<std::streamsize>::max()) {
+        throw std::runtime_error("ensure_fixed_contents: string is too large");
+    }
+
+    std::string actual = read_bytes(std::streamsize(len));
 
     if (actual != expected) {
         // NOTE: I think printing it outright is not best idea, it could contain non-ascii charactes like backspace and beeps and whatnot. It would be better to print hexlified version, and also to redirect it to stderr.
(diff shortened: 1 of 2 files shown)


kaitai/kaitaistream.cpp
```diff
@@ -112,30 +112,30 @@ int8_t kaitai::kstream::read_s1() {
...
 
 int16_t kaitai::kstream::read_s2be() {
-    int16_t t;
+    uint16_t t;
     m_io->read(reinterpret_cast<char *>(&t), 2);
 #if __BYTE_ORDER == __LITTLE_ENDIAN
     t = bswap_16(t);
 #endif
-    return t;
+    return int16_t(t);
```
GreyCat: Can you explain what's the purpose of this change?

kaitai/kaitaistream.cpp
```diff
@@ -251,20 +251,24 @@ uint64_t kaitai::kstream::read_u8le() {
...
 float kaitai::kstream::read_f4be() {
     uint32_t t;
+    float out;
     m_io->read(reinterpret_cast<char *>(&t), 4);
 #if __BYTE_ORDER == __LITTLE_ENDIAN
     t = bswap_32(t);
 #endif
-    return reinterpret_cast<float&>(t);
+    type_pun(out, t);
+    return out;
```
GreyCat: Can you explain why `type_pun` was actually needed, and what's wrong with `reinterpret_cast` solution? `type_pun` implementation seems to do a call to `std::memcpy` — I'm not sure what's the purpose of that and if that ever gets optimized out, as we don't really need any memory copying here, it's just a matter of reinterpreting existing memory contents (likely in RAX, not even on stack).

kaitai/kaitaistream.cpp
maks500: `bswap_{16,32,64}` takes unsigned int, the function returns signed int which causes a sign conversion warning.

kaitai/kaitaistream.cpp
maks500: `type_pun` is just a utility. `reinterpret_cast` in this case [violates strict aliasing rules](https://github.com/CppCon/CppCon2017/raw/master/Presentations/Type%20Punning%20In%20C%2B%2B17%20-%20Avoiding%20Pun-defined%20Behavior/Type%20Punning%20In%20C%2B%2B17%20-%20Avoiding%20Pun-defined%20Behavior%20-%20Scott%20Schurr%20-%20CppCon%202017.pdf) (Page 70). The call to `std::memcpy` is optimized, doesn't violate aliasing, and the code produced is the same as if using `reinterpret_cast`.

GreyCat:
Merging this one in, thanks for this massive effort!

kaitai/kaitaistream.h
```diff
@@ -261,6 +270,25 @@ class kstream {
...
+template <class From>
+//inline typename std::make_signed<From>::type to_signed(From from) {
+inline auto to_signed(From from) {
+    using signed_F = typename std::make_signed<From>::type;
+
+    if (from <= std::numeric_limits<signed_F>::max()) {
+        return static_cast<signed_F>(from);
+    }
+
+    throw std::runtime_error("toSigned: Invalid casting");
```
generalmimon: @GreyCat @maks500 
Unfortunately, this line is currently causing the whole C++11 CI build on Linux to fail (https://travis-ci.org/kaitai-io/ci_targets/jobs/660553811#L584):

```cpp
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_11/../../../runtime/cpp_stl/kaitai/kaitaistream.h:282:11: error: ‘runtime_error’ is not a member of ‘std’
     throw std::runtime_error("toSigned: Invalid casting");
```

Another issue is that it's a bad idea to throw generic runtime exceptions, it's generally recomended to always create specific exceptions which can be handled much easier from the application code.

webbnh: > ‘runtime_error’ is not a member of ‘std’

Actually, it _is_ a member of `std`, so perhaps there's a missing header file or similar issue?

> it's a bad idea to throw generic runtime exceptions

I would suggest that that depends on the situation.  If the intention is that the caller be able to recover from the exception, then the I concur that the API should include defined exception types or objects which convey to the caller all the information required to handle the situation when one is thrown; however, if the expectation is that there is no recovery available, then there is likely no reason to create anything fancier than one of the "standard" exception objects -- all that is required is that a human be able to diagnose the problem after the fact.  (As such, I'd say that this code looks about right, although it might be nice if the message included a little more detail, such as the location in the KSY where the offending cast can be found.)

generalmimon: > however, if the expectation is that there is no recovery available, then there is likely no reason to create anything fancier than one of the "standard" exception objects -- all that is required is that a human be able to diagnose the problem after the fact.

I agree, but if this is the case, then I'd say it's not a valid use case for `runtime_error` from the semantic POV. There are generally two groups of exceptions: **runtime** and **logic** ones. Runtime errors can be found only on runtime occurs, whereas logic errors _represent error in the program logic and are _presumably detectable before the program executes_ ([source](http://www.cplusplus.com/reference/stdexcept/logic_error/)). That means that logic errors are not intended to be caught by the application code, so there is no point creating some specific subclasses.

So if this error has logic nature (as you denote), then it should be [`std::logic_error`](http://www.cplusplus.com/reference/stdexcept/logic_error/). But  **runtime errors** (for example end-of-stream errors are typical runtime ones - you can't detect them by just inspecting the code because they can be triggered just by some specific user input) should IMHO **always** have specific subclasses, because they're intended to be caught by the application code.

webbnh: I suspect that we are going to be challenged by terminology and context here.  :-)

I believe that there are in fact _two_ run times in play:  there is the operation of the Kaitai Struct Compiler, and there is the operation of the code that it produced.  Concordantly, there are two sets of "user input", respectively:  the KSY file, and the to-be-parsed data.

If the exception is raised during the execution of KSC in response to some aspect of the KSY, then I think `std::runtime_error` is reasonable.  If the exception is raised during the execution of the resulting application's call to the KSC-generated code, then things are murkier.  (I apologize for not having a good enough understanding of this code change to be able to tell where we are....)

In the latter case, I don't particularly like the use of `std::runtime_error`, but I think it's acceptable; however, I don't particularly like the use of `std::logic_error` either.  What we're trying to report is, presumably, an error which occurred during the construction of the application due to a feature of the KSY, but which wasn't detected until a execution of the application.  In any case, there is nothing that the application can do about it at run-time, and so I don't think a subclass (etc.) is warranted.

(And, I disagree with your assertion that just because something is a "runtime error" means that it is something that the application can or is intended to be able to recover from -- some errors are "fatal" -- but they can and should be reported via exceptions to allow the callers to restore any external invariants prior to halting execution.  If they are detected at run-time and don't represent an error in the internal logic of the program, there is no reason not to report them as `std::runtime_error`'s.)

generalmimon:
@maks500 
I added the missing `#include <stdexcept>` in commit https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/commit/3e1556057995d42015dd1a0d82879a5778c8aff6 to restore [C++11 CI build on Linux](https://ci.kaitai.io/). However, changes in this PR apparently use some features introduced in C++11 and not available in C++98, so it broke all C++98 CI builds in the CI.

For the record, this says the _clang_ compiler about it (https://travis-ci.org/github/kaitai-io/ci_targets/jobs/661709006#L582):

```cpp
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:276:8: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
inline auto to_signed(From from) {
       ^
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:276:8: error: 'auto' not allowed in function return type
inline auto to_signed(From from) {
       ^~~~
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:277:22: warning: alias declarations are a C++11 extension [-Wc++11-extensions]
    using signed_F = typename std::make_signed<From>::type;
                     ^
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:277:36: error: no type named 'make_signed' in namespace 'std'
    using signed_F = typename std::make_signed<From>::type;
                     ~~~~~~~~~~~~~~^~~~~~~~~~~
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:277:47: error: expected ';' after alias declaration
    using signed_F = typename std::make_signed<From>::type;
                                              ^
                                              ;
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:279:37: error: use of undeclared identifier 'signed_F'; did you mean 'signed'?
    if (from <= std::numeric_limits<signed_F>::max()) {
                                    ^~~~~~~~
                                    signed
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:280:28: error: unknown type name 'signed_F'; did you mean 'signed'?
        return static_cast<signed_F>(from);
                           ^~~~~~~~
                           signed
/home/travis/build/kaitai-io/ci_targets/tests/spec/cpp_stl_98/../../../runtime/cpp_stl/kaitai/kaitaistream.h:289:5: error: use of undeclared identifier 'static_assert'; did you mean 'static_cast'?
    static_assert(sizeof(read) == sizeof(output), "Type sizes don't match");
    ^
/home/travis/build/kaitai-io/ci_targets/runtime/cpp_stl/kaitai/kaitaistream.cpp:570:5: error: use of undeclared identifier 'static_assert'; did you mean 'static_cast'?
    static_assert(sizeof(std::streamoff) <= sizeof(uint64_t), "pos_to_uint: invalid conversion");
    ^
2 warnings and 7 errors generated.
```

@GreyCat 
I would consider reverting this breaking PR until the issue is resolved, as I fear that the support for C++98 will fade out as in the case of C#. It's really great that we're halfway to [C# Async support](https://github.com/kaitai-io/kaitai_struct/issues/640) for the price of broken CI build, but it would IMHO be much better if it at least worked properly in the CI.

GreyCat:
Unfortunately, I have to agree — let's revert it for now and take another closer look on how to reimplement it in a way that will work for all C++ environments.

PR #36, opened 2020-04-26
Add method read_bits_int_le

generalmimon:
See kaitai-io/kaitai_struct#155

I'm not much proficient in C++, so I don't dare commit it without approval. It's basically the same code [as used in Java runtime](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/6f9fbf949188a0e13a3c7a1524a5a7c4bc316c5c/src/main/java/io/kaitai/struct/KaitaiStream.java#L233-L256).


generalmimon:
@GreyCat Thanks for the review, I'm merging it then.

PR #37, opened 2020-09-10
cmake: add header dir to target

russkel:
This simplifies use of the runtime in CMake projects. With this mod it's now possible to simply do:

```cmake
add_subdirectory(external/kaitai_struct_cpp_stl_runtime)

add_executable(myproj main.cpp etcetcetc)
target_link_libraries(myproj PRIVATE
  etc etc etc et
  kaitai_struct_cpp_stl_runtime
)
```
Before I was getting this error, due to it not finding the includes correctly.

```
In file included from ksy/myproj_protocol.cpp:3:
ksy/myproj_protocol.h:6:10: fatal error: kaitai/kaitaistruct.h: No such file or directory
    6 | #include "kaitai/kaitaistruct.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[2/18] Building CXX object src/CMakeFiles/myproj.dir/myproj.cpp.o
FAILED: src/CMakeFiles/myproj.dir/myproj.cpp.o 
/usr/bin/c++  -I../external/PcapPlusPlus/Dist/header -Iksy -I../ -g -std=c++17 -MD -MT src/CMakeFiles/myproj.dir/myproj.cpp.o -MF src/CMakeFiles/myproj.dir/myproj.cpp.o.d -o src/CMakeFiles/myproj.dir/myproj.cpp.o -c ../src/myproj.cpp
../src/myproj.cpp:5:10: fatal error: kaitai/kaitaistream.h: No such file or directory
    5 | #include <kaitai/kaitaistream.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```


LHLaurini:
Would you mind rebasing this pull request? Maybe we can get a maintainer to merge it.

CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(STRING_ENCODING_TYPE "ICONV" CACHE STRING "Set the way strings have to be en
 
 add_library (${PROJECT_NAME} SHARED ${HEADERS} ${SOURCES})
 set_target_properties(${PROJECT_NAME} PROPERTIES PUBLIC_HEADER "kaitai/kaitaistream.h;kaitai/kaitaistruct.h")
+target_include_directories(${PROJECT_NAME} INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
```
generalmimon, line 31: Can you explain why you believe that [`CMAKE_CURRENT_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html) is the right variable to use here?

Some people use [`CMAKE_CURRENT_LIST_DIR`](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_LIST_DIR.html) in this context instead, see https://stackoverflow.com/a/40242257/12940655. And the official CMake tutorial uses [`PROJECT_SOURCE_DIR`](https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html), see https://cmake.org/cmake/help/latest/guide/tutorial/Adding%20a%20Library.html#build-and-run:

> Finally we need to specify the library's header file location. Modify [**`target_include_directories()`**](https://cmake.org/cmake/help/latest/command/target_include_directories.html#command:target_include_directories) to add the `MathFunctions` subdirectory as an include directory so that the `MathFunctions.h` header file can be found.
>
> _TODO 4: CMakeLists.txt_
> ```cmake
> target_include_directories(Tutorial PUBLIC
>                           "${PROJECT_BINARY_DIR}"
>                           "${PROJECT_SOURCE_DIR}/MathFunctions"
>                           )
> ```

Intuitively, `PROJECT_SOURCE_DIR` sounds most predictable and robust to me and it's probably what we actually want (given that the official tutorial uses it too). The word "current" in `CMAKE_CURRENT_*` variable names suggests that the contents of these variables may change under some circumstances (which might be some other CMake way to include `kaitai_struct_cpp_stl_runtime` to someone else's application) and that could break things.

russkel: Honestly I have never seen `PROJECT_SOURCE_DIR` in the wild. If it works then great.

russkel:
> Would you mind rebasing this pull request? Maybe we can get a maintainer to merge it.

I don't have anything to test it on anymore given this was 3 years ago.

CMakeLists.txt
LHLaurini: In this case, all three are equivalent. How I understand it:

- `CMAKE_CURRENT_SOURCE_DIR` is the source directory being processed. It is "current" because that can change with `add_subdirectory`;
- `CMAKE_CURRENT_LIST_DIR` is the directory containing the current CMake file. In this case, the listing is in the source dir;
- `PROJECT_SOURCE_DIR` is the source directory from the last call to `project`.

So, IMO using either `CMAKE_CURRENT_SOURCE_DIR` or `PROJECT_SOURCE_DIR` would be sensible (and equivalent) choices, but I do agree that `PROJECT_SOURCE_DIR` may be clearer.

generalmimon (approved):
@russkel @LHLaurini Thank you both!

PR #40, opened 2021-06-11
Add public headers installation for WIN32

### LINKED ISSUE #41: CMake doesn't install public headers on Windows
#16 introduced VC++ support to the library and added branching for shared library installation. Then, 817568db4c715b3228df917e8489de722c117753 added public header installation but for some reason it was missed for `WIN32`.

I resolved it in #40. This issue blocks adding the library to conan (https://github.com/conan-io/conan-center-index/pull/5132) so this minor change may help the good deal.

If I didn't understand reasons for the status quo, feel free to close #40 and let me know if this issue should be resolved in another way.

N-Nagorny:
This PR adds public headers installation for WIN32 as it's already done for other platforms.

Resolves #41


generalmimon:
Cc @webbnh

N-Nagorny:
Another thing to do is letting Microsoft linker find symbols. The library may be static for Windows and shared for other platforms (few lines in CMakeLists.txt) or it may stay shared but in this case all public functions should be [exported](https://docs.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport) to build the library and [imported](https://docs.microsoft.com/en-us/cpp/build/importing-into-an-application-using-declspec-dllimport) to build an app (modifying all function declarations).

What approach is more preferrable to you?

ams-tschoening:
@N-Nagorny Please make more clear what exactly your second commit does: From my understanding, it does not change shared to static, but fixes/improves the shared build. Am I wrong? I'm wondering a bit which [problem issued for conan](https://github.com/conan-io/conan-center-index/pull/5132#issuecomment-813665849) you are actually addressing. the shared vs. static thing didn't sound like it triggered build errors to me, but it actually did and that's why you propose your changes?

A change/discussion about shared to static should be an individual PR instead in my opinion. If things are shared right now, kept and this PR is needed to make it properly work with conan, merging looks fine to me. OTOH, static vs. shared in general is a broader topic and should not block the fix necessary for conan in my opinion.

N-Nagorny:
@ams-tschoening, right, it does not change shared to static after all. [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) creates a `.lib` file containing symbols, `ARCHIVE` installs it to the package to let the test program be linked with the `.dll`.

Regarding Conan, despite [this](https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#kb-h050-default-shared-option-value) it did the trick. I built the package locally with the code from this PR despite it has only the shared version and the test package links the library dynamically. I didn't try to build a user app using this package, maybe it won't be the same as building the program from the test package.

ams-tschoening (approved):

CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ set (SOURCES
 
 set(STRING_ENCODING_TYPE "ICONV" CACHE STRING "Set the way strings have to be encoded (ICONV|NONE|...)")
 
+if (WIN32)
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
```
generalmimon, line 31: Is it necessary to wrap this setting into the `if (WIN32)` condition? I usually see `CMakeLists.txt` without it (https://github.com/robotology/how-to-export-cpp-library/blob/afb21efb655e7b1cc11636c3d42aad9e02fb626f/CMakeLists.txt#L48), but not sure if it doesn't fail under non-Windows systems in that case.

CMakeLists.txt
```diff
@@ -50,7 +54,10 @@ endif()
 
 if (WIN32)
     install(TARGETS ${PROJECT_NAME}
-        RUNTIME DESTINATION lib)
+        ARCHIVE DESTINATION lib
+        RUNTIME DESTINATION bin
+        PUBLIC_HEADER DESTINATION include/kaitai
+    )
 else()
```
generalmimon, line 61: Does this branching to Windows and non-Windows systems actually serve any purpose? I mean, why not just

```diff
-if (WIN32)
     install(TARGETS ${PROJECT_NAME}
         ARCHIVE DESTINATION lib
         RUNTIME DESTINATION bin
+        LIBRARY DESTINATION lib
         PUBLIC_HEADER DESTINATION include/kaitai
     )
-else()
-    install (TARGETS ${PROJECT_NAME}
-        LIBRARY DESTINATION lib
-        PUBLIC_HEADER DESTINATION include/kaitai
-    )
-endif()
```

...? The result will be the same, right?

CMakeLists.txt
N-Nagorny: It is not necessary and it doesn't fail under Linux for instance. I've added the `if` to make it explicit for those who read `CMakeLists.txt` but I fixed it in the suggestion as you proposed.

```suggestion
set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```

CMakeLists.txt
N-Nagorny: Yes, the result is the same. I preferred a more explicit way in order to make cross-platform support maintenance simpler. Do you want me to change that to your snippet? I don't mind it, just let me know.

ams-tschoening: Am I correct that ARCHIVE, RUNTIME and LIBRARY are simply set to their default values? If so, you might want to discuss removing them in favour of making explicit that PUBLIC_HEADER is the only non-default setting changed for a reason. Though, explicitly documenting the used paths might have a benefit as well of course.

https://cmake.org/cmake/help/latest/command/install.html#installing-targets

CMakeLists.txt
```diff
@@ -50,7 +54,10 @@ endif()
 
 if (WIN32)
     install(TARGETS ${PROJECT_NAME}
-        RUNTIME DESTINATION lib)
+        ARCHIVE DESTINATION lib
+        RUNTIME DESTINATION bin
+        PUBLIC_HEADER DESTINATION include/kaitai
```
GreyCat, line 59: I'm very hesitating about adding it like that. All the best practices normally dictate using `#include <library_name/header_name.h>`, not `#include <header_name.h>` to avoid collisions. Adding `kaitai` directly would promote bad habits and will make it possible to get super weird stuff like `#include "exceptions.h"` randomly including Kaitai Struct exceptions, not your applications one.

ams-tschoening: As I understood things, `PUBLIC_HEADER` in this case only configures where to store headers of the built lib and is not an include dir:

> Any PUBLIC_HEADER files associated with a library are **installed in the destination specified** by the PUBLIC_HEADER argument on non-Apple platforms.[...]

https://cmake.org/cmake/help/latest/command/install.html#targets

So other projects would exactly need to do what you require from my understanding, because the created file path would be `[...]/include/kaitai/header_name.h`:

    #include <kaitai/header_name.h>

Without that statement from my understanding things would result in exactly the behaviour you don't want, because the created file path would be `[...]/include/header_name.h` by default.

N-Nagorny:
I added commits obeying the PR comments and merged `master` to this branch to get it usable until this PR is merged.

N-Nagorny:
Any other changes required?

CMakeLists.txt
GreyCat: Got it, thanks for the detailed explanation!

generalmimon:
@N-Nagorny

> Any other changes required?

If you tested the latest `CMakeLists.txt` version in practice and it works (**please confirm so that we can proceed**), I don't see any serious obstacle to merging. I'm just thinking that seeing these placeholder keys without any settings looks a bit weird...

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/5977ba116e6ee7c2f0bbcc20f29718b1420bdfc9/CMakeLists.txt#L54-L57

There seems to be no reason why these dummy lines are there, it's rather confusing. I'd either set all default destinations explicitly again so that these lines have some reason to be there, or remove them completely. Users would be then expected to know that the CMake's default values apply in that case (https://cmake.org/cmake/help/latest/command/install.html#installing-targets):

> The following table shows the target types with their associated variables and built-in defaults that apply when no destination is given:
> 
> Target Type | GNUInstallDirs Variable | Built-In Default
> -- | -- | --
> `RUNTIME` | `${CMAKE_INSTALL_BINDIR}` | `bin`
> `LIBRARY` | `${CMAKE_INSTALL_LIBDIR}` | `lib`
> `ARCHIVE` | `${CMAKE_INSTALL_LIBDIR}` | `lib`
> `PRIVATE_HEADER` | `${CMAKE_INSTALL_INCLUDEDIR}` | `include`
> `PUBLIC_HEADER` | `${CMAKE_INSTALL_INCLUDEDIR}` | `include`

or we may point that out in the README or a CMakeLists.txt comment or whatever, if users would have problems with that. The dummy lines do not serve that purpose, actually they don't serve _any_ good purpose for that matter.

ams-tschoening:
> There seems to be no reason why these dummy lines are there, it's rather confusing. I'd either set all default destinations explicitly again so that these lines have some reason to be there, or remove them completely.[...]

Things heavily depend on if without `ARCHIVE` etc. at all CMAKE still does what `ARCHIVE` etc. should trigger. I had the feeling it does and therefore suggested to remove those lines entirely, as configuring default values again doesn't make too much sense to me.

After @N-Nagorny only removed the paths, I had the impression that the keywords themself need to be present to trigger CMAKE to actually do `ARCHIVE` etc. Though, I didn't want to discuss this concrete change further because it wasn't that important at all.

But in the end, I agree with @generalmimon: If the keywords need to be present to trigger CMAKE doing things, provide a path to be in line with `PUBLIC_HEADER`. If the keywords do not need to be present at all, but CMAKE still does `ARCHIVE`, remove the lines without paths as well, because it really looks somewhat confusing. Additional docs or whatever don't make things any better.

Sorry for bringing this up in the first place, was totally unnecessary! :-/

N-Nagorny:
The keywords are necessary to trigger CMake, so I turned back the explicit default paths. I also checked the last commit on a Linux system (with and without Conan) and a Windows one (with Conan): the builds passed without errors.

CMakeLists.txt
```diff
@@ -49,12 +51,9 @@ else()
...
-else()
-    install (TARGETS ${PROJECT_NAME}
-        LIBRARY DESTINATION lib
-        PUBLIC_HEADER DESTINATION include/kaitai
-    )
-endif()
+install(TARGETS ${PROJECT_NAME}
+    ARCHIVE DESTINATION lib
+    RUNTIME DESTINATION bin
+    LIBRARY DESTINATION lib
```
generalmimon, line 57: My personal taste tells me to swap these lines, so I'll do it:

```suggestion
    ARCHIVE DESTINATION lib
    LIBRARY DESTINATION lib
    RUNTIME DESTINATION bin
```

generalmimon (approved):
LGTM, thanks for the contribution! And apologies for the delay.

PR #43, opened 2021-06-14
Fix kstream constructor which uses std::string for initialization

N-Nagorny:
https://en.cppreference.com/w/cpp/io/basic_istringstream/basic_istringstream (3) takes `const std::string&` so it looks appropriate to take the same in `kaitai::kstream` constructor working with `std::string`.

Resolves https://github.com/kaitai-io/kaitai_struct/issues/806


generalmimon (approved):
Thanks!

PR #54, opened 2023-02-10
Add QNX support

jayoh-dev:
Add support for QNX


GreyCat:
Sounds great! I don't think we realistically can test it routinely in CI, do I'll take your word that it works. It shouldn't affect other platforms. Merging.

jayoh-dev:
@GreyCat Thanks for the consideration!
I'll keep watch your great project and check under QNX works.

The preprocessor definition `__QNX__` is from qcc, the QNX compiler, following:
http://www.qnx.com/developers/docs/qnxcar2/index.jsp?topic=%2Fcom.qnx.doc.neutrino.prog%2Ftopic%2Fdevel_Neutrino_specific.html

PR #56, opened 2023-06-02
GitHub Actions CI for running unit tests

GreyCat:
* Adds GitHub Actions-based CI workflows to build library + unit tests and do a basic test run.
* Implements 2x2 scripts in `.build/` for Linux/Windows and build/run steps.
* Adds CI workflows for Linux and Windows.
* Fixes one warning in .h file (problem proven by [CI failing run](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/actions/runs/5154249448/jobs/9282479655), successful unit tests prove that nothing has been broken after the fix).

Known caveats:

* Tests rely on availability of GTest, but different OSes have different ways to install and versions to offer:
  * For Linux, it's installed in CI using apt (takes ~15-20s). As of now, this gets *GTest 1.11.0-3*.
  * For Windows, it's more complicated: there's no binary build, so we utilize vcpkg to download and build gtest during CI (slightly longer: takes ~50-110s). As of now, this gets *GTest 1.13.0*.
* Versions of OS/compilers are not fixed and might change at will by GitHub. As of now, they are:
  * Linux
    * Ubuntu 22.04.2 LTS
    * GCC 11.3.0
  * Windows:
    * Microsoft Windows Server 2022 Datacenter, 10.0.20348
    * MSVC 19.35.32217.1

### DIFF
--- a/.build/build
+++ b/.build/build
@@ -1,0 +1,8 @@
+#!/bin/sh -ef
+
+cd $(dirname "$0")/..
+
+mkdir -p build
+cd build
+cmake ..
+cmake --build .
--- a/.build/build
+++ b/.build/build
@@ -1,6 +1,6 @@
 #!/bin/sh -ef
 
-cd $(dirname "$0")/..
+cd "$(dirname "$0")"/..
 
 mkdir -p build
 cd build
--- a/.build/build.ps1
+++ b/.build/build.ps1
@@ -10,6 +10,7 @@ Requires:
 
 [CmdletBinding()]
 param (
+    [Parameter(Mandatory=$true)]
     [string] $GTestPath
 )
 
@@ -18,8 +19,12 @@ Set-StrictMode -Version Latest
 $ErrorActionPreference = "Stop"
 $PSDefaultParameterValues['*:ErrorAction']='Stop'
 
-$null = New-Item ..\build -ItemType Directory -Force
-Push-Location ..\build
+# Go to repo root
+$repoRoot = (Resolve-Path "$PSScriptRoot\..").Path
+Push-Location $repoRoot
+
+$null = New-Item build -ItemType Directory -Force
+cd build
 
 cmake -DCMAKE_PREFIX_PATH="$GTestPath" -DSTRING_ENCODING_TYPE=NONE ..
 cmake --build . --config Debug
--- a/.build/build.ps1
+++ b/.build/build.ps1
@@ -1,0 +1,28 @@
+<#
+.DESCRIPTION
+Builds Kaitai Struct C++ runtime library and unit tests
+
+Requires:
+- MSVC native tools installed and available in the command prompt
+- cmake/ctest available (normally installed with MSVC native tools)
+- GTest installed, path passed in `-GTestPath`
+#>
+
+[CmdletBinding()]
+param (
+    [string] $GTestPath
+)
+
+# Standard boilerplate
+Set-StrictMode -Version Latest
+$ErrorActionPreference = "Stop"
+$PSDefaultParameterValues['*:ErrorAction']='Stop'
+
+Push-Location ..\build
+
+cmake -DCMAKE_PREFIX_PATH="$GTestPath" -DSTRING_ENCODING_TYPE=NONE ..
+cmake --build . --config Debug
+cp $GTestPath\debug\bin\*.dll tests\Debug
+cp Debug\kaitai_struct_cpp_stl_runtime.dll tests\Debug
+
+Pop-Location
--- a/.build/build.ps1
+++ b/.build/build.ps1
@@ -18,6 +18,7 @@ Set-StrictMode -Version Latest
 $ErrorActionPreference = "Stop"
 $PSDefaultParameterValues['*:ErrorAction']='Stop'
 
+$null = New-Item ..\build -ItemType Directory -Force
 Push-Location ..\build
 
 cmake -DCMAKE_PREFIX_PATH="$GTestPath" -DSTRING_ENCODING_TYPE=NONE ..
--- a/.build/run-unittest
+++ b/.build/run-unittest
@@ -1,6 +1,6 @@
 #!/bin/sh -ef
 
-cd $(dirname "$0")/..
+cd "$(dirname "$0")"/..
 
 cd build/tests
 ./unittest
--- a/.build/run-unittest
+++ b/.build/run-unittest
@@ -1,0 +1,6 @@
+#!/bin/sh -ef
+
+cd $(dirname "$0")/..
+
+cd build/tests
+./unittest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,3 +18,17 @@ jobs:
         run: .build/build
       - name: unittest
         run: .build/run-unittest
+
+  windows:
+    runs-on: windows-latest
+    # NB: see https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md for what's available in this image
+    steps:
+      - uses: actions/checkout@v1
+      - name: restore
+        run: |
+          C:\vcpkg\bootstrap-vcpkg.bat
+          C:\vcpkg\vcpkg install gtest
+      - name: build
+        run: .build\build.ps1 -GTestPath C:\vcpkg\installed\x64-windows
+      - name: unittest
+        run: .build\run-unittest.ps1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,20 @@
+name: Unit tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request: {}
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: restore
+        run: |
+          apt-get install -y libgtest-dev
+      - name: build
+        run: .build/build
+      - name: unittest
+        run: .build/run-unittest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
       - name: restore
         run: |
           C:\vcpkg\bootstrap-vcpkg.bat
-          C:\vcpkg\vcpkg install gtest
+          C:\vcpkg\vcpkg install gtest:x64-windows
       - name: build
         run: .build\build.ps1 -GTestPath C:\vcpkg\installed\x64-windows
       - name: unittest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v1
       - name: restore
         run: |
-          apt-get install -y libgtest-dev
+          sudo apt-get install -y libgtest-dev
       - name: build
         run: .build/build
       - name: unittest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: windows-latest
     # NB: see https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md for what's available in this image
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
       - name: restore
         run: |
           C:\vcpkg\bootstrap-vcpkg.bat
--- a/kaitai/kaitaistream.h
+++ b/kaitai/kaitaistream.h
@@ -245,8 +245,40 @@ public:
         char buf[std::numeric_limits<I>::digits10 + 5];
         if (val < 0) {
             buf[0] = '-';
-            // get absolute value without undefined behavior (https://stackoverflow.com/a/12231604)
-            unsigned_to_decimal(-static_cast<int64_t>(val), &buf[1]);
+
+            // NB: `val` is negative and we need to get its absolute value (i.e. minus `val`). However, since
+            // `int64_t` uses two's complement representation, its range is `[-2**63, 2**63 - 1] =
+            // [-0x8000_0000_0000_0000, 0x7fff_ffff_ffff_ffff]` (both ends inclusive) and thus the naive
+            // `-val` operation will overflow for `val = std::numeric_limits<int64_t>::min() =
+            // -0x8000_0000_0000_0000` (because the result of `-val` is mathematically
+            // `-(-0x8000_0000_0000_0000) = 0x8000_0000_0000_0000`, but the `int64_t` type can represent at
+            // most `0x7fff_ffff_ffff_ffff`). And signed integer overflow is undefined behavior in C++.
+            //
+            // To avoid undefined behavior for `val = -0x8000_0000_0000_0000 = -2**63`, we do the following
+            // steps for all negative `val`s:
+            //
+            // 1. Convert the signed (and negative) `val` to an unsigned `uint64_t` type. This is a
+            //    well-defined operation in C++: the resulting `uint64_t` value will be `val mod 2**64` (`mod`
+            //    is modulo). The maximum `val` we can have here is `-1` (because `val < 0`), a theoretical
+            //    minimum we are able to support would be `-2**64 + 1 = -0xffff_ffff_ffff_ffff` (even though
+            //    in practice the widest standard type is `int64_t` with the minimum of `-2**63`):
+            //
+            //    * `static_cast<uint64_t>(-1) = -1 mod 2**64 = 2**64 + (-1) = 0xffff_ffff_ffff_ffff = 2**64 - 1`
+            //    * `static_cast<uint64_t>(-2**64 + 1) = (-2**64 + 1) mod 2**64 = 2**64 + (-2**64 + 1) = 1`
+            //
+            // 2. Subtract `static_cast<uint64_t>(val)` from `2**64 - 1 = 0xffff_ffff_ffff_ffff`. Since
+            //    `static_cast<uint64_t>(val)` is in range `[1, 2**64 - 1]` (see step 1), the result of this
+            //    subtraction will be mathematically in range `[0, (2**64 - 1) - 1] = [0, 2**64 - 2]`. So the
+            //    mathematical result cannot be negative, hence this unsigned integer subtraction can never
+            //    wrap around (which wouldn't be a good thing to rely upon because it confuses programmers and
+            //    code analysis tools).
+            //
+            // 3. Since we did mathematically `(2**64 - 1) - (2**64 + val) = -val - 1` so far (and we wanted
+            //    to do `-val`), we add `1` to correct that. From step 2 we know that the result of `-val - 1`
+            //    is in range `[0, 2**64 - 2]`, so adding `1` will not wrap (at most we could get `2**64 - 1 =
+            //    0xffff_ffff_ffff_ffff`, which is still in the valid range of `uint64_t`).
+
+            unsigned_to_decimal((std::numeric_limits<uint64_t>::max() - static_cast<uint64_t>(val)) + 1, &buf[1]);
         } else {
             unsigned_to_decimal(val, buf);
         }
--- a/kaitai/kaitaistream.h
+++ b/kaitai/kaitaistream.h
@@ -246,7 +246,7 @@ public:
         if (val < 0) {
             buf[0] = '-';
             // get absolute value without undefined behavior (https://stackoverflow.com/a/12231604)
-            unsigned_to_decimal(-static_cast<uint64_t>(val), &buf[1]);
+            unsigned_to_decimal(-static_cast<int64_t>(val), &buf[1]);
         } else {
             unsigned_to_decimal(val, buf);
         }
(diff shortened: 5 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
...
+  push:
+    branches:
+      - master
+  pull_request: {}
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
generalmimon, line 13:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+          sudo apt-get install -y libgtest-dev
+      - name: build
+        run: .build/build
+      - name: unittest
+        run: .build/run-unittest
+
+  windows:
+    runs-on: windows-latest
+    # NB: see https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md for what's available in this image
```
generalmimon, line 24: I don't like seeing links to the moving `main`/`master` branch for any reason, can you please choose a commit hash or tag that's permanent?

generalmimon (changes requested):

generalmimon:
@GreyCat:

> I'll try to follow up with bringing up an actual C++98 compiler in CI to make sure we are making sense of C++98 compatibility mode. As of now, I'm fairly sure we use tons of C++11 features despite claiming we're C++98 compatible.

Hmm, you think despite [`spec/cpp_stl_98/CMakeLists.txt:14`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac51f82a98e3425b1f9ffa7a75375ec4ac6acc12/spec/cpp_stl_98/CMakeLists.txt#L14):

```cpp
set(CMAKE_CXX_STANDARD 98)
```

... a modern C++ compiler may still allow some post-C++98 features?

generalmimon:
@GreyCat Also, if you absolutely can't stand this in the code:

```cpp
            unsigned_to_decimal(-static_cast<uint64_t>(val), &buf[1]);
```

we can rewrite it like this:

```cpp
            unsigned_to_decimal(0 - static_cast<uint64_t>(val), &buf[1]);
```

and because of the modular arithmetic, that can be equivalently written as:

```cpp
            unsigned_to_decimal((std::numeric_limits<uint64_t>::max() - static_cast<uint64_t>(val)) + 1, &buf[1]);
```

And hopefully this might also silence the MSVC's warning, because we're using a binary minus instead of unary minus, so perhaps it starts believing that we know what we're doing.

generalmimon:
@GreyCat Also feel free to experiment with the different methods at **https://godbolt.org/z/d3Th3b88x**:

image

I've included `-fsanitize=undefined` in compile options for both `clang` and `gcc` - as you can see, `abs_old` and `abs_new` don't cause any undefined behavior known to the UB sanitizers, whereas the obvious way of negating an `int64_t` in `abs_wrong` does.

kaitai/kaitaistream.h
```diff
@@ -246,7 +246,7 @@ class kstream {
         if (val < 0) {
             buf[0] = '-';
             // get absolute value without undefined behavior (https://stackoverflow.com/a/12231604)
-            unsigned_to_decimal(-static_cast<uint64_t>(val), &buf[1]);
+            unsigned_to_decimal(-static_cast<int64_t>(val), &buf[1]);
         } else {
```
generalmimon, line 250: @GreyCat To summarize, I'd recommend this:

```cpp
        if (val < 0) {
            buf[0] = '-';

            // NB: `val` is negative and we need to get its absolute value (i.e. minus `val`). However, if `int64_t`
            // uses two's complement representation (which is most common), then its range is `[-2**63, 2**63 - 1] =
            // [-0x8000_0000_0000_0000, 0x7fff_ffff_ffff_ffff]` (both ends inclusive) and thus the naive `-val`
            // operation will overflow for `val = std::numeric_limits<int64_t>::min() = -0x8000_0000_0000_0000` (because
            // the result of `-val` is mathematically `-(-0x8000_0000_0000_0000) = 0x8000_0000_0000_0000`, but the
            // `int64_t` type can represent at most `0x7fff_ffff_ffff_ffff`). And signed integer overflow is undefined
            // behavior in C++.
            //
            // To avoid undefined behavior for `val = -0x8000_0000_0000_0000 = -2**63`, we do the following steps for
            // all negative `val`s:
            //
            // 1. Convert the signed (and negative) `val` to an unsigned `uint64_t` type. This is a well-defined
            //    operation in C++: the resulting `uint64_t` value will be `val mod 2**64` (`mod` is modulo). The
            //    maximum `val` we can have here is `-1` (because `val < 0`), a theoretical minimum we are able to
            //    support would be `-2**64 + 1 = -0xffff_ffff_ffff_ffff` (even though in practice the only standard type
            //    is `int64_t` with the minimum of `-2**63`):
            //
            //    * `static_cast<uint64_t>(-1) = -1 mod 2**64 = 2**64 + (-1) = 0xffff_ffff_ffff_ffff = 2**64 - 1`
            //    * `static_cast<uint64_t>(-2**64 + 1) = (-2**64 + 1) mod 2**64 = 2**64 + (-2**64 + 1) = 1`
            //
            // 2. Subtract `static_cast<uint64_t>(val)` from `2**64 - 1 = 0xffff_ffff_ffff_ffff`. Since
            //    `static_cast<uint64_t>(val)` is in range `[1, 2**64 - 1]` (see step 1), the result of this subtraction
            //    will be mathematically in range `[0, (2**64 - 1) - 1] = [0, 2**64 - 2]`. So the mathematical result
            //    cannot be negative, hence this unsigned integer subtraction can never wrap around from negative to
            //    positive values (which wouldn't be a good thing to rely upon because it confuses programmers and code
            //    analysis tools).
            //
            // 3. Since we did mathematically `(2**64 - 1) - (2**64 + val) = -val - 1` so far (and we wanted to do
            //    `-val`), we add `1` to correct that. From step 2 we know that the result of `-val - 1` is in range
            //    `[0, 2**64 - 2]`, so adding `1` will not wrap (at most we could get `2**64 - 1 =
            //    0xffff_ffff_ffff_ffff`, which is still in the valid range of `uint64_t`).

            unsigned_to_decimal((std::numeric_limits<uint64_t>::max() - static_cast<uint64_t>(val)) + 1, &buf[1]);
        } else {
            unsigned_to_decimal(val, buf);
        }
```

Let me know if anything is unclear now.

GreyCat:
> > I'll try to follow up with bringing up an actual C++98 compiler in CI to make sure we are making sense of C++98 compatibility mode. As of now, I'm fairly sure we use tons of C++11 features despite claiming we're C++98 compatible.
> 
> Hmm, you think that despite [`spec/cpp_stl_98/CMakeLists.txt:14`](https://github.com/kaitai-io/kaitai_struct_tests/blob/ac51f82a98e3425b1f9ffa7a75375ec4ac6acc12/spec/cpp_stl_98/CMakeLists.txt#L14):
> 
> ```c++
> set(CMAKE_CXX_STANDARD 98)
> ```
> 
> ... a modern C++ compiler may still allow some post-C++98 features?

First of all, this actually is not enforced/checked in CMakeLists.txt in this repo. Secondly, yeah, I believe this only tweaks *some* compiler behaviors (especially those which have changed), but largely doesn't influence stdlibs and those are not very specific about additional features it enables. Finally, some of the stuff we use is actually not C++98, but, for example, C99.

Here's, for example, what happens when I try to build this repo with gcc 4.1.3 from 2006:

```
$ gcc --version
gcc (GCC) 4.1.3 20070929 (prerelease) (Ubuntu 4.1.2-16ubuntu2)
Copyright (C) 2006 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ make
[100%] Building CXX object CMakeFiles/kaitai_struct_cpp_stl_runtime.dir/kaitai/kaitaistream.o
/kaitai_struct/runtime/cpp_stl/kaitai/kaitaistream.cpp: In member function 'uint64_t kaitai::kstream::read_bits_int_be(int)':
/kaitai_struct/runtime/cpp_stl/kaitai/kaitaistream.cpp:334: error: 'UINT64_C' was not declared in this scope
/kaitai_struct/runtime/cpp_stl/kaitai/kaitaistream.cpp: In member function 'uint64_t kaitai::kstream::read_bits_int_le(int)':
/kaitai_struct/runtime/cpp_stl/kaitai/kaitaistream.cpp:378: error: 'UINT64_C' was not declared in this scope
make[2]: *** [CMakeFiles/kaitai_struct_cpp_stl_runtime.dir/kaitai/kaitaistream.o] Error 1
make[1]: *** [CMakeFiles/kaitai_struct_cpp_stl_runtime.dir/all] Error 2
make: *** [all] Error 2
```

UINT64_C is only available as C99 feature, and if we really care about true C++98 support, we shouldn't really rely on that. Instead of `UINT64_C(1)`, probably `static_cast<uint64_t>(1)` will be a safer choice.

I'll try to set up separate CI stage to track that and a separate issue/PR to propose a fix.

.github/workflows/build.yml
GreyCat: The caveat here is that `runs-on: windows-latest` is a moving target, and that moving link with `main` kind of matches that to some effect.

Even if we'll fix it as `runs-on: windows-2022`, it will still change arbitrarily by GitHub, and, sadly, having that non-permanent link is what will describe it best.

generalmimon: I get that, but my experience says that generally it's very likely that a link to `main` branch will entirely stop working in a year, for example (because repos get often restructured, so what has a certain filename/path now is likely to be renamed/relocated somewhere else entirely). So you get no content instead of at least some content (and with that, it's easier to find where the latest version is if you need it than if you get a Not Found page).

But OK, I try to keep an eye on whether the links still work once in a while, so I'll take care of this one too.

generalmimon:
@GreyCat:

> UINT64_C is only available as C99 feature, and if we really care about true C++98 support, we shouldn't really rely on that. Instead of `UINT64_C(1)`, probably `static_cast<uint64_t>(1)` will be a safer choice.

Hmm, I didn't think of that. Yeah, now I see that actually all [fixed width integer types](https://en.cppreference.com/w/cpp/types/integer) were introduced in C++11.

> # Fixed width integer types (since C++11)

This is also mentioned in this blog post [**Pitfalls in Writing Portable C++98 Code**](https://kevinushey.github.io/blog/2016/09/14/pitfalls-in-writing-portable-c++98-code/#c99):

> ### C99
> 
> The C++98 standard was ratified at an unfortunate time – it came just one year before the C99 standard, and the C99 standard introduced a number of tools that make it easier to write safe + correct C code. Some of the newer pieces that became part of the C99 standard made it into C++98, but some didn’t. Examples that I’ve bumped into thus far are:
> 
> - `long long`
> - `snprintf` / `vsnprintf`
> - `isblank`
> - <ins>**Fixed-width integer types (`uint8_t` etc., from [`cstdint`](http://en.cppreference.com/w/cpp/types/integer))**</ins>
> - Variadic macros

---

Indeed, that looks pretty bad for the C++98 compatibility - do you think we can avoid using all the `uint64_t`-like types entirely? I'm not sure whether it's even possible for our use case, because the standard integer types do not guarantee much (they can have various sizes and representations on different platforms, etc.).

For example, when we're reading a 64-bit signed integer in `read_s8le`, we badly need an integer type that's (at least) 64 bits in size and has 2's complement representation. `int64_t` is such a type, but I have no idea how do you get away without it.

generalmimon:
It seems that https://www.boost.org/doc/libs/1_82_0/libs/config/doc/html/boost_config/cstdint.html could help with the fixed width integer types in C++98. Although I'm not sure whether it's worth it, achieving C++98 compatibility is apparently much harder than I thought...

.build/build.ps1
```diff
@@ -0,0 +1,34 @@
...
+    [string] $GTestPath
+)
+
+# Standard boilerplate
+Set-StrictMode -Version Latest
+$ErrorActionPreference = "Stop"
+$PSDefaultParameterValues['*:ErrorAction']='Stop'
+
+# Go to repo root
+$repoRoot = (Resolve-Path "$PSScriptRoot\..").Path
```
generalmimon, line 23: If I see correctly, this doesn't go to the repo root. In GitHub Actions workflows, the default working directory is already the repo root (and the `.github/workflows/build.yml` workflow doesn't say otherwise), and mere invocation of a script as `.build\build.ps1` doesn't change the working directory (and the script is also evaluated in the context of the repo root). So if you go to the parent directory anyway, you end up in a parent directory of the repo root, whatever that is.

generalmimon: Sorry, I got it wrong here. Still, I think specifying paths relative to the script location like this is confusing and adds unnecessary complexity to all scripts. In general, programs usually do not have this behavior of hardcoding paths of stuff they're working with relative to where they're located. Much more common (and also predictable and flexible) is to use the current working directory to expect certain files/paths to exist (for example `Makefile`). Or, even more flexible and explicit is to simply require the paths to work with to be passed as command-line arguments (I don't personally think we need it here, but it's arguably the clearest option).

I see that your motive probably was to make these two usages equivalent:

```bash
PS C:\temp\kaitai_struct\runtime\cpp_stl> .\.build\build.ps1
```

```bash
PS C:\temp\kaitai_struct\runtime\cpp_stl> cd .build
PS C:\temp\kaitai_struct\runtime\cpp_stl\.build> .\build.ps1
```

but I don't think it's needed. As far as I'm concerned, just expecting that the script will be run from the repo root directory is OK and also easy to understand.

kaitai/kaitaistream.h
GreyCat: @generalmimon Thanks very much for such a thorough investigation and a comment. Let me give it try.

.build/build.ps1
GreyCat: You're correct about the purpose, and, ultimately, it's absolutely a matter of personal preference.

I'm obviously biased, but from my experience I've found that having in-repo scripts which work without relying on current working directory being set to anything specific is less error-prone and raises fewer questions — both from humans and from various CI implementations which might want to start at some specific location or limit your ability to just launch one script rather than do `cd somewhere && path/to/script`.

To certain extent, you probably see that 75% of the point of existence of this script in the first place is cmake CLI interface that is indeed highly dependent on current working dir, forcing you to do these weird invocations like `cd build; cmake ..`

If you strongly oppose that, sure, I'll remove it.

GreyCat:
@generalmimon Sorry about late responses to this proposal.

Let me also share my general view on what I want to achieve with us stating "C++xx compatible". I personally don't care that much about achieving certain theoretical standard, but what I do care about is people actually being able to use our software without major hindrances. If standards help that, that's great. If they don't, I would always favor reproducing and resolving certain real-life issues over blindly following the standard and trying to fight compilers.

I'm well aware how shaky C++ standards are — e.g. whole [codecvt](https://en.cppreference.com/w/cpp/locale/codecvt) is fairly buggy and being almost unusable in gcc/libstdc++ up until 2015-2016, despite being part of C++98 standard.

So, my rule of thumb is reproducing the problem (ideally, in CI) and offering a fix to it, rather than trying to fight abstractly for standard purity.

.build/build.ps1
generalmimon: > If you strongly oppose that, sure, I'll remove it.

My stance on this is actually quite neutral; you're right that it comes down to personal preference. From my standpoint, it's just an unnecessary operation in the code that may potentially contain a bug - which in the case of the shell script it actually does, so let me point it out...

.build/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -ef
+
+cd $(dirname "$0")/..
```
generalmimon, line 3: See https://www.shellcheck.net/

<pre><code>Line 3:
cd $(dirname "$0")/..
   ^-- <a href="https://www.shellcheck.net/wiki/SC2046">SC2046</a> (warning): Quote this to prevent word splitting.</code></pre>

```suggestion
cd "$(dirname "$0")"/..
```

.build/run-unittest
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -ef
+
+cd $(dirname "$0")/..
```
generalmimon, line 3:
```suggestion
cd "$(dirname "$0")"/..
```

generalmimon (changes requested):

.build/build
GreyCat: Actually, shellcheck seems to be wrong here. There are 2 shells:

* Internal one executes `dirname "$0"`, and this one indeed requires quoting, otherwise `$0` gets expanded first and might result in more than 1 argument being sent over to dirname.
* External one executes `cd ` + result from internal shell + `/..`. Parsing and tokenization of external one actually gets decided first (otherwise, you won't be able to even derive that there is internal shell call), so it's already been decided that there's just 2 tokens: 1st is `cd` and 2nd will be decided later as result of execution of that expression.

So, there's no real need to quote it.

Quick test: I'm in the following path: `/home/greycat/z/x  y/a b` (note two spaces between `x  y` and one space between `a b`).

```
$ pwd | hd   
00000000  2f 68 6f 6d 65 2f 67 72  65 79 63 61 74 2f 7a 2f  |/home/greycat/z/|
00000010  78 20 20 79 2f 61 20 62  0a                       |x  y/a b.|
$ cat foo.sh
#!/bin/sh -e

pwd
cd "$(dirname "$0")"/..
pwd
$ ./foo.sh
/home/greycat/z/x  y/a b
/home/greycat/z/x  y
```

For sanity checking, just checked it with dash (=sh), bash, zsh.

generalmimon: > Actually, shellcheck seems to be wrong here.

I don't think so.

> So, there's no real need to quote it.
> 
> Quick test: I'm in the following path: `/home/greycat/z/x y/a b` (note two spaces between `x y` and one space between `a b`).

This test doesn't show anything for 2 reasons - according to the `$ cat foo.sh` output that you did there, you were already testing the corrected solution with quoted `"$(dirname "$0")"`, so no wonder you can't find any issues with it.

The other reason becomes obvious when you actually print the contents of the `$0` variable and the result of `dirname "$0"`, for example like this (and also notice that now I've actually removed the quotes in `cd $(dirname "$0")`, so we'll be able to see how it's wrong):

```console
pp@DESKTOP-89OPGF3:~/x y/a b$ cat foo.sh
#!/bin/sh -e

printf %s "$0" | hd
printf %s "$(dirname "$0")" | hd

pwd
cd $(dirname "$0")/..
pwd
```

If you run the script as `./foo.sh`:

```console
pp@DESKTOP-89OPGF3:~/x y/a b$ ./foo.sh
00000000  2e 2f 66 6f 6f 2e 73 68                           |./foo.sh|
00000008
00000000  2e                                                |.|
00000001
/home/pp/x y/a b
/home/pp/x y
```

You can see that the `$0` contained exactly the path `./foo.sh` that the script was run with, and `dirname "$0"` is `.` (just 1-character string, the dot) - since `.` doesn't contain any spaces, the unquoted version indeed works in this case.

However, if the invocation path includes a directory name with a space, the unquoted version stops working (as expected; this is a shell script where you really have to quote almost everything, I didn't make it up):

```console
pp@DESKTOP-89OPGF3:~/x y$ ./a\ b/foo.sh
00000000  2e 2f 61 20 62 2f 66 6f  6f 2e 73 68              |./a b/foo.sh|
0000000c
00000000  2e 2f 61 20 62                                    |./a b|
00000005
/home/pp/x y
./a b/foo.sh: 7: cd: can't cd to ./a
```

GreyCat: Uhh, you're absolutely correct, sorry for all the confusion. My bad for not noticing it. Let me apply the fix.

GreyCat:
@generalmimon I think this is more or less ready to merge after application of fixes you've suggested? Can I ask you to take one last look and sign off? Once ready, I'll rebase #57 and we can get some degree of relatively ancient C++98 support validated.

generalmimon (approved):
@GreyCat Looks good to me, thanks!

Feel free to merge once you're ready.

.github/workflows/build.yml
generalmimon: > I get that, but my experience says that generally it's very likely that a link to `main` branch will stop working in a year

@GreyCat Just noting that I was right - https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md is now dead. It's actually dead for more than a year now, because it was moved in commit https://github.com/actions/runner-images/commit/d1f2c9a3bedac344533bd2499133f030adab7d90#diff-d3119a278ce9ce80fc4b5af63a1112fb3a044aec3e6002d838e8eb26d7a3a499 (authored on Nov 15, 2023) to `images/windows/Windows2022-Readme.md`.

GreyCat: That's genuinely sad :(

I guess the only saving grace is that it's still a git repo and it's possible to get this file history tracked...

But I agree with you, modern software development front seem to rot faster than anything...

PR #57, opened 2023-06-16
Add CI support for gcc 4.1.2 / Ubuntu 7.10

GreyCat:
This add CI support for ancient Ubuntu 7.10 and ancient gcc 4.1.2. This PR is based off https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/56, and it supposed to be reviewed after that.

Ubuntu 7.10 ships with:

```
gcc (GCC) 4.1.3 20070929 (prerelease) (Ubuntu 4.1.2-16ubuntu2)
Copyright (C) 2006 Free Software Foundation, Inc.

cmake version 2.4-patch 7
```

## Challenges

This one is challenging:

* It's impossible to run git clone inside this container (as old git would use older protocols not supported anymore by GitHub, and also standard `actions/checkout@v3` will depend on NodeJS/npm, which are not available in this container). To work around this, we start/stop container manually and use targeted execution of specific build/run commands only inside the container, keeping git and networking stuff out.

* For gcc, vast majority of C++11 features are missing from this release (many were added around GCC 4.3 and 4.4). Some C99 features are also missing. C++98 support seems to be full-featured.

* Modern GTest doesn't work in this environment, as GTest v1.12.x requires C++11 and modern GTest v1.13+ requires C++14. To work around this, we supply nano substitute of GTest (tests/gtest-nano.h), partially compatible with API. The only difference is we need to swap `#include <gtest/gtest.h>` to something different, so I've added GTEST_NANO define to distinguish between these two options.

* CMake is way too old to be usable without major rewrite of all the CMakeLists.txt to use ancient/deprecated syntax. To work around this, we use plain GNU make(1) and supply custom Makefile.gcc4.

## Problems found and fixed along with this implementation

* UINT64_C used in kaitaistream.cpp is not available in this gcc, as this is part of C99, not C++98 (see [this CI report for failure](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/actions/runs/5291489434/jobs/9577107282)). Fixed by replacing with static_cast.

## Extra stuff

Using custom test implementation also requires minimal testing. Here's an example of results of what happens when we introduce error in the test expectations:

* Ubuntu 22.04 (with regular GTest): https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/actions/runs/5291541904/jobs/9577225385
* Ubuntu 7.10 (with "GTest" nano): https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/actions/runs/5291541904/jobs/9577225574

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,11 +51,11 @@ jobs:
           docker pull icomputer7/ancient-ubuntu-docker:gutsy
       - name: docker create start
         run: |
-          docker create --name u710 --label u710 --workdir /__w/kaitai_struct_cpp_stl_runtime/kaitai_struct_cpp_stl_runtime -v "${{ github.workspace }}":"/__w" --entrypoint "tail" icomputer7/ancient-ubuntu-docker:gutsy "-f" "/dev/null"
+          docker create --name u710 --label u710 --workdir /w -v "${{ github.workspace }}":"/w" --entrypoint "tail" icomputer7/ancient-ubuntu-docker:gutsy "-f" "/dev/null"
           docker start u710
       - name: restore
         run: |
-          docker exec u710 sh -c 'apt-get update && apt-get -y --no-install-recommends install make g++'
+          docker exec u710 sh -c 'pwd && ls -al && apt-get update && apt-get -y --no-install-recommends install make g++'
       - name: build
         run: |
           docker exec u710 sh -c 'make -f Makefile.gcc4'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,21 +39,23 @@ jobs:
   #   To work around this, we use plain GNU make(1) and supply custom Makefile.gcc4.
   linux_ubuntu_7_10:
     runs-on: ubuntu-latest
+    container: 'icomputer7/ancient-ubuntu-docker:gutsy'
     steps:
-      - uses: actions/checkout@v3
       - name: restore
-        container: 'icomputer7/ancient-ubuntu-docker:gutsy'
         run: |
           apt-get update
           apt-get -y --no-install-recommends install \
               git \
               make \
               g++
+      # Can't use actions/checkout@v3, as it depends on NodeJS, which is not available in this image
+      - name: checkout
+        run: |
+          git clone ${{ github.repositoryUrl }} .
+          git checkout ${{ github.sha }}
       - name: build
-        container: 'icomputer7/ancient-ubuntu-docker:gutsy'
         run: make -f Makefile.gcc4
       - name: unittest
-        container: 'icomputer7/ancient-ubuntu-docker:gutsy'
         run: build/unittest
 
   windows:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
           docker pull icomputer7/ancient-ubuntu-docker:gutsy
       - name: docker create start
         run: |
-          docker create --name u710 --label u710 --workdir /w -v "${{ github.workspace }}":"/w" --entrypoint "tail" icomputer7/ancient-ubuntu-docker:gutsy "-f" "/dev/null"
+          docker create --name u710 --label u710 --workdir /w -v "$GITHUB_WORKSPACE":"/w" --entrypoint "tail" icomputer7/ancient-ubuntu-docker:gutsy "-f" "/dev/null"
           docker start u710
       - name: restore
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,40 @@ jobs:
       - name: unittest
         run: .build/run-unittest
 
+  # Builds and runs tests on Ubuntu 7.10.
+  # It ships with:
+  #
+  # gcc (GCC) 4.1.3 20070929 (prerelease) (Ubuntu 4.1.2-16ubuntu2)
+  # Copyright (C) 2006 Free Software Foundation, Inc.
+  #
+  # cmake version 2.4-patch 7
+  #
+  # This one is challenging:
+  #
+  # * For gcc, vast majority of C++11 features are missing from this release (many were added around GCC 4.3 and 4.4).
+  #   Some C99 features are also missing. C++98 support seems to be full-featued.
+  #
+  # * Modern GTest doesn't work in this environment, as GTest v1.12.x requires C++11 and modern GTest v1.13+ requires C++14.
+  #   To work around this, we supply nano substitute of GTest (tests/gtest-nano.h), partially compatible with API.
+  #
+  # * CMake is way too old to be usable without major rewrite of all the CMakeLists.txt to use ancient/deprecated syntax.
+  #   To work around this, we use plain GNU make(1) and supply custom Makefile.gcc4.
+  linux_ubuntu_7_10:
+    runs-on: icomputer7/ancient-ubuntu-docker:gutsy
+    steps:
+      - name: restore
+        run: |
+          sudo apt-get update
+          sudo apt-get -y --no-install-recommends install \
+              git \
+              make \
+              g++
+      - uses: actions/checkout@v3
+      - name: build
+        run: make -f Makefile.gcc4
+      - name: unittest
+        run: build/unittest
+
   windows:
     runs-on: windows-latest
     # NB: see https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md for what's available in this image
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,9 @@ jobs:
   # * CMake is way too old to be usable without major rewrite of all the CMakeLists.txt to use ancient/deprecated syntax.
   #   To work around this, we use plain GNU make(1) and supply custom Makefile.gcc4.
   linux_ubuntu_7_10:
-    runs-on: icomputer7/ancient-ubuntu-docker:gutsy
+    runs-on: ubuntu-latest
+    container:
+      image: icomputer7/ancient-ubuntu-docker:gutsy
     steps:
       - name: restore
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,9 @@ jobs:
       # Can't use actions/checkout@v3, as it depends on NodeJS, which is not available in this image
       - name: checkout
         run: |
-          git clone ${{ github.repositoryUrl }} .
+          git clone ${{ github.repositoryUrl }} __repo
+          find __repo -mindepth 1 -maxdepth 1 -print0 | xargs -0 -i mv '{}' .
+          rmdir __repo
           git checkout ${{ github.sha }}
       - name: build
         run: make -f Makefile.gcc4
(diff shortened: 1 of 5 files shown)


Makefile.gcc4
```diff
@@ -0,0 +1,33 @@
...
+	mkdir -p $(BUILD_DIR)
+
+$(BUILD_DIR)/unittest: $(OBJS)
+	$(CXX) $(CXXFLAGS) -o build/unittest $(OBJS) $(LFLAGS) $(LIBS)
+
+$(BUILD_DIR)/kaitaistream.o: kaitai/kaitaistream.cpp
+	$(CXX) $(CXXFLAGS) $(INCLUDES) $(DEFINES) -c $< -o $@
+
+$(BUILD_DIR)/unittest.o: tests/unittest.cpp tests/gtest-nano.h
+	$(CXX) $(CXXFLAGS) $(INCLUDES) $(DEFINES) -c $< -o $@
```
generalmimon, line 33: Is there a reason you specify `tests/gtest-nano.h` as a dependency of `unittest.o`, but not `kaitai/kaitaistream.h` which both object files depend on?

```suggestion
$(BUILD_DIR)/kaitaistream.o: kaitai/kaitaistream.cpp kaitai/kaitaistream.h
	$(CXX) $(CXXFLAGS) $(INCLUDES) $(DEFINES) -c $< -o $@

$(BUILD_DIR)/unittest.o: tests/unittest.cpp tests/gtest-nano.h kaitai/kaitaistream.h
	$(CXX) $(CXXFLAGS) $(INCLUDES) $(DEFINES) -c $< -o $@
```

Technically, if the `kaitai/kaitaistream.h` file is modified, both object files have to be recompiled.

.github/workflows/build.yml
```diff
@@ -19,6 +19,53 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: docker pull
+        run: |
+          docker pull icomputer7/ancient-ubuntu-docker:gutsy
+      - name: docker create start
+        run: |
+          docker create --name u710 --label u710 --workdir /w -v "${{ github.workspace }}":"/w" --entrypoint "tail" icomputer7/ancient-ubuntu-docker:gutsy "-f" "/dev/null"
+          docker start u710
```
generalmimon, line 55: I don't like the way `${{ github.workspace }}` is injected there. You're letting the GitHub runner to interpolate the default working directory to some text that will be later used as a shell command, so the runner is clueless about shell syntax and what are the characters with special meaning for the shell in the double quoted string literal (`$`, `` ` ``, `\`, `"` and possibly others) that must not appear in the `github.workspace`. Maybe in this particular case we don't expect that the directory GitHub came up will be this crazy, but to me it's wrong anyway (and I generally also find it easier and much less error-prone to only ever use the safe and reliable patterns than having to analyze every time whether the unsafe pattern is "safe enough" for a given use case or not).

The simplest solution I see is to let the shell do the variable expansion and use the `GITHUB_WORKSPACE` environment variable instead (see [Default environment variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)):

```suggestion
      - name: docker create start
        run: |
          docker create --name u710 --label u710 --workdir /w -v "$GITHUB_WORKSPACE":"/w" --entrypoint "tail" icomputer7/ancient-ubuntu-docker:gutsy "-f" "/dev/null"
          docker start u710
```

tests/gtest-nano.h
```diff
@@ -0,0 +1,66 @@
+// gtest-nano.h implements very minimalistic GTest-compatible API that can be used to run tests in older
+// (C++98-compatible) environments.
+
+#include <iostream>
+#include <vector>
+
+namespace testing {
+    struct TestInfo {
+        const char* suite;
```
generalmimon, line 9: Generally, I strongly recommend moving the `*` (for pointers) and `&` (for references) closer to the variable name because of how the C/C++ semantics work when declaring multiple variables (this is also the reason I took the liberty of reformatting the existing code with that in mind in commit https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/commit/aaa2a3ff7ce5981d9ab1f3ada36e6633c5e96162).

```diff
-        const char* suite;
+        const char *suite;
```

This convention allows you to correctly predict what will be the result of the following code snippets:

```cpp
int foo(void) {
    int *a = 0, b = 0;
    return a == 0 ? a : b;
}
// test.cpp:3:19: error: incompatible operand types ('int *' and 'int')
//     return a == 0 ? a : b;
//                   ^ ~   ~
```

(☝️ And note that `clang` also prefers to print the type as `'int *'` in the error message, in line with my preferences.)

```cpp
#include <iostream>
int main() {
    std::string s = "Changed";
    std::string &x = s, y = s;

    x += " through x";
    y += " through y";

    std::cout << s << std::endl; // => "Changed through x"
}
```

I mean, we're just humans, so code formatting sometimes plays a big role how we may (mis)interpret things, and if you write <code><del>int*</del> a = 0, b = 0</code> (or <code><del>std::string&</del> x = s, y = s</code>) instead (which I don't recommend), it's easy to get fooled and think that both `a`, `b` have the type <code><del>int*</del></code> (which is not the case) etc.

Makefile.gcc4
GreyCat: Nope, just forgot about that. Thanks for spotting!

.github/workflows/build.yml
GreyCat: Makes sense, let's give it a try. In all fairness, though, I believe you're right and GitHub will never use anything which needs complicated quoting in this place.

tests/gtest-nano.h
GreyCat: I've missed the change, thanks for pointing out. At the very least this should be consistent with the rest of the codebase.

We can have separate discussion on pros and cons of both of these approaches. From my personal experience (although I'm biased, as a human being), in the latest years `int* a` is a more popular one than `int *a`, and double declaration on the same command line is normally banned in the style guide and checker.

GreyCat: I believe it's done, please take one final look?

Makefile.gcc4
```diff
@@ -0,0 +1,33 @@
...
+
+INCLUDES := -I.
+
+all: $(BUILD_DIR) $(BUILD_DIR)/unittest
+
+$(BUILD_DIR):
+	mkdir -p $(BUILD_DIR)
+
+$(BUILD_DIR)/unittest: $(OBJS)
+	$(CXX) $(CXXFLAGS) -o build/unittest $(OBJS) $(LFLAGS) $(LIBS)
```
generalmimon, line 27: Technically, the standard names (used also in implicit rules of GNU make) for what you call _`LFLAGS`_ and _`LIBS`_ would be `LDFLAGS` and `LDLIBS` (see https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html#index-linking_002c-predefined-rule-for) - see also https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-LDFLAGS. Of course, such a convention is not that important, but I guess it's good to follow.

GreyCat: Thanks for the catch, just fixed!

generalmimon (approved):
@GreyCat Looks good, thanks for your work!

Makefile.gcc4
```diff
@@ -0,0 +1,33 @@
...
+
+LDFLAGS :=
+LDLIBS :=
+
+INCLUDES := -I.
+
+all: $(BUILD_DIR) $(BUILD_DIR)/unittest
+
+$(BUILD_DIR):
+	mkdir -p $(BUILD_DIR)
```
generalmimon, line 24: Actually, I'm a little bit unsure whether this "directory rule" will work correctly in `make` (and whether `make` will understand that this `$(BUILD_DIR)` rule in fact is a prerequisite to of all the `$(BUILD_DIR)/...` stuff). I wonder whether it shouldn't be added as a dependency of the `$(BUILD_DIR)/...` rules in some way.

It seems that people on the internet do the same stuff as we do here, and [in this SO answer](https://stackoverflow.com/a/4251368/12940655) someone suggests that GNU make has a concept of [order-only prerequisites](http://www.gnu.org/s/hello/manual/make/Prerequisite-Types.html) that are supposedly appropriate for this case, but I've never heard of it.

To me, the easiest workaround that avoids all these questions would be to have the `build/` directory tracked by Git again - I see you removed it from Git index in https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/commit/aa6e59b556b451674964e2ecf4a88ec73b2b454e, which I personally have no problem with, but it would've avoided these problems.

Makefile.gcc4
```diff
@@ -0,0 +1,33 @@
...
+	$(BUILD_DIR)/unittest.o
+
+DEFINES := -DKS_STR_ENCODING_ICONV -DGTEST_NANO
+
+LDFLAGS :=
+LDLIBS :=
+
+INCLUDES := -I.
+
+all: $(BUILD_DIR) $(BUILD_DIR)/unittest
```
generalmimon, line 21: Yeah, and `all` should ideally be a [phony target](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html), but that's a trifle:

```suggestion
.PHONY: all
all: $(BUILD_DIR) $(BUILD_DIR)/unittest
```

Makefile.gcc4
GreyCat: Sharing my point of view on this: in general, this whole Makefile-something is a hack, which exists only to prove certain degree of compatibility with ancient compilers. It doesn't have to be particularly robust or be well-tested in all possible scenarios.

I completely understand that Makefiles per se are a pretty dangerous conversation to have. You start with something simple (2-3 two-liner rules), and then you start generalizing, you start supporting all possible cleanups and permutations of partial builds, and several weeks after you realize that your Makefile is incomprehensible mess of 1000s lines of code with lots of obscure syntax and ifdefs. And we haven't even started touching the topic of GNU Make vs BSD Make, etc.

This is pretty much exactly the reason why vast majority of even 1990s C/C++ projects strive away from Makefiles altogether, at the very least abstracting this part out with autotools or cmake or anything else.

In this specific case, I would leave it as is, as long as it moderately good does its only job: proving that KS C++ runtime works on Ubuntu 7.10.

Having directories like `build/` added to version control is generally a bad idea — if anything, for sole reason of being able to deleting whole `build/` to reset and not worry about version control crying about lost gitignore and/or placeholder file. If a build process generates `build/tests/Debug/...`, why shouldn't it create `build/` in the first place?

generalmimon: > Sharing my point of view on this: in general, this whole Makefile-something is a hack, which exists only to prove certain degree of compatibility with ancient compilers. It doesn't have to be particularly robust or be well-tested in all possible scenarios.

Fair enough, I'm completely OK with this. Probably I tend to overgeneralize things, but I see the benefit of writing code that is safe to use not only in the one place for which it was intended, but also safe for inspiration when there's a need to do a similar thing for another purpose in the future, without having to analyze the differences of situations and why it may have been OK to use a pattern in one place but not in this other place.

Makefile.gcc4
GreyCat: Cool, let me add that: f653068dc553f7eb452b26b5e71f2a861b065690

Makefile.gcc4
GreyCat: Totally get your aspiration, but for this specific case I would say that we should rather discourage people from building on top of ancient compilers :)

PR #61, opened 2023-07-12
Add Win32API-based encoding option to C++ runtime

GreyCat:
This adds capability to convert string encodings in KS runtime C++ on Windows using only system-supplied Win32 APIs. This offers alternative to bringing external iconv library (which is the only way we support string conversion in C++ now).

Other changes worth mentioning:

* This assumes standardized encoding names, as runtime now does conversion from encoding name in string format into Windows codepage (which is an integer).
  * Actual discussion around standard for encoding names is happening in https://github.com/kaitai-io/kaitai_struct/issues/116, and there's a table there which I've started filling out to decide on canonical encoding names.
  * In future, ksc will make sure that runtime always gets passed a canonical name, and will either issue a warning or an error to the user for specifying non-canonical name in their .ksy files.
  * Adjusted the tests to match these now-canonical names.
* This conversion (string->integer) is costly (and mostly unnecessary to do in runtime), so in future I will offer a way to do convert encoding string in compile-time.
* To be able to do that, I've started switching from `std::string src_enc` to `const char* src_enc`, which paves the way to optimize it further.
* Added UTF-16BE test.
* Reformatted most of the encoding tests to spell out exact codepoints which are returned.


generalmimon:
@GreyCat:

> * This assumes standardized encoding names, as runtime now does conversion from encoding name in string format into Windows codepage (which is an integer).

Well... why not - given that `STRING_ENCODING_TYPE = "WIN32API"` is a new feature, it can provide partial support, but without accompanying compiler changes to canonicalize the encoding names, this implementation will be considerably stricter than the vast majority of target languages we support. What I have in mind is that the `encoding` key should for example:

* be case-insensitive (e.g. if `ASCII` works, then `ascii` or `Ascii` should too, with or without eventual warnings)
* tolerate typographical mistakes in hyphens (i.e. not only canonical spellings like `UTF-8` and `UTF-16BE` should be accepted, but also `utf8`, `utf_8`, `UTF16BE`, `UTF16-BE` or `UTF-16-BE`, etc.)
* as you know, sometimes there are several distinct similarly popular names for one encoding, e.g. `Shift_JIS` vs. `SJIS`, etc.

I support the idea of having a clear idea what encodings Kaitai Struct supports and what their canonical names are (https://github.com/kaitai-io/kaitai_struct/issues/393#issuecomment-1632535399), but I think we should do it in reverse order (first have the compiler canonicalize, then we can potentially take advantage of accepting only canonical names in runtime libraries).

GreyCat:
@generalmimon That's exactly what I've started with, but then I've realized that implementing all that in one single branch in C++ is roughly same amount of work as implementing this in the compiler, so I went ahead and done https://github.com/kaitai-io/kaitai_struct_compiler/pull/254. Please take a look too?

kaitai/kaitaistream.cpp
```diff
@@ -635,9 +635,10 @@ uint8_t kaitai::kstream::byte_array_max(const std::string val) {
 #include <iconv.h>
 #include <cerrno>
 #include <stdexcept>
+#include "kaitaistream.h"
```
generalmimon, line 638: Umm..., what? This should have been obviously already included at the very beginning of `kaitai/kaitaistream.cpp`:

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/2939b29cc38962c931d29087bf0a6872aa767bd6/kaitai/kaitaistream.cpp#L1

So I see no point in doing it for the second time (`kaitaistream.h` has header guards, so it won't have any effect, but... why).

kaitai/kaitaistream.cpp
```diff
@@ -635,9 +635,10 @@ uint8_t kaitai::kstream::byte_array_max(const std::string val) {
...
 
-std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc) {
-    iconv_t cd = iconv_open(KS_STR_DEFAULT_ENCODING, src_enc.c_str());
+std::string kaitai::kstream::bytes_to_str(const std::string src, const char* src_enc) {
```
generalmimon, line 640: Again, I'd prefer this a bit more (see our previous discussion in https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/57#discussion_r1233123710):

```diff
-std::string kaitai::kstream::bytes_to_str(const std::string src, const char* src_enc) {
+std::string kaitai::kstream::bytes_to_str(const std::string src, const char *src_enc) {
```

...but this is really more like a job for an autoformatter.

kaitai/kaitaistream.cpp
```diff
@@ -691,9 +694,143 @@ std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc)
...
-std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc) {
+std::string kaitai::kstream::bytes_to_str(const std::string src, const char* src_enc) {
     return src;
 }
+#elif defined(KS_STR_ENCODING_WIN32API)
+#include <windows.h>
+#include <limits>
+
+// Unbreak std::numeric_limits<T>::max, as otherwise MSVC substitutes "useful" max() macro.
+#undef max
```
generalmimon, line 705: What a mess 🤦‍♂️

Given the circumstances, I guess this workaround is OK, though.

kaitai/kaitaistream.cpp
```diff
@@ -691,9 +694,143 @@ std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc)
...
+        // If our source is already UTF16LE, just copy it
+        utf16_len = src_len / 2;
+        utf16 = std::wstring((wchar_t*)src.c_str(), utf16_len);
+        break;
+    case KAITAI_CP_UTF16BE:
+        // If our source is in UTF16BE, convert it to UTF16LE by swapping bytes
+        utf16_len = src_len / 2;
+
+        utf16 = std::wstring(utf16_len, L'\0');
+        for (int i = 0; i < utf16_len; i++) {
```
generalmimon, line 799: In practice, `int` is perhaps always the same as `int32_t` _on Windows_, so it won't make much of a difference, but I suppose `i` should be rather `size_t` or at least `int32_t` (which is apparently enough in this case because `src_len` is at most `std::numeric_limits<int32_t>::max()` and `utf16_len = src_len / 2`).

kaitai/kaitaistream.cpp
GreyCat: Oops, sorry, looks like some "smart" IDE tried to help and added it here, and I've overlooked it. Will fix.

kaitai/kaitaistream.cpp
GreyCat: Fixed, thanks for the reminder. We should definitely set up some clang-format or something like that.

kaitai/kaitaistream.cpp
GreyCat: It is :( I'm yet to see a non-trivial Windows C++ project which doesn't do a bunch of tweaks like that (e.g. undefining min/max, defining _UNICODE, etc), sadly.

kaitai/kaitaistream.cpp
GreyCat: Good catch! Let me at least make it int32_t everywhere.

tests/unittest.cpp
```diff
@@ -67,29 +67,61 @@ TEST(KaitaiStreamTest, bytes_to_str_ascii)
 }
 
 #ifndef KS_STR_ENCODING_NONE
-TEST(KaitaiStreamTest, bytes_to_str_iso8859_1)
+TEST(KaitaiStreamTest, bytes_to_str_iso_8859_1)
 {
-    std::string res = kaitai::kstream::bytes_to_str("\xC4\xD6\xDC\xE4\xF6\xFC\xDF", "ISO8859-1");
-    EXPECT_EQ(res, "\xC3\x84\xC3\x96\xC3\x9C\xC3\xA4\xC3\xB6\xC3\xBC\xC3\x9F");
+    std::string res = kaitai::kstream::bytes_to_str("\xC4\xD6\xDC\xE4\xF6\xFC\xDF", "ISO-8859-1");
```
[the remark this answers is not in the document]
GreyCat: Ok, makes sense. Let me add 0xa6, which will become U+0160 `Š` in ISO-8859-15, and it will be U+00A6 `¦` in ISO-8859-1.

tests/unittest.cpp
```diff
@@ -67,29 +67,77 @@ TEST(KaitaiStreamTest, bytes_to_str_ascii)
...
+    std::string res = kaitai::kstream::bytes_to_str("\xC4\xD6\xDC\xE4\xF6\xFC\xDF\xA6", "ISO-8859-1");
+    EXPECT_EQ(res,
+        "\xC3\x84"  // U+00C4 LATIN CAPITAL LETTER A WITH DIAERESIS
+        "\xC3\x96"  // U+00D6 LATIN CAPITAL LETTER O WITH DIAERESIS
+        "\xC3\x9C"  // U+00DC LATIN CAPITAL LETTER U WITH DIAERESIS
+        "\xC3\xA4"  // U+00E4 LATIN SMALL LETTER A WITH DIAERESIS
+        "\xC3\xB6"  // U+00F6 LATIN SMALL LETTER O WITH DIAERESIS
+        "\xC3\xBC"  // U+00FC LATIN SMALL LETTER U WITH DIAERESIS
+        "\xC3\x9F"  // U+00DF LATIN SMALL LETTER SHARP S
+        "\xC2\xA6"  // U+00A6 BROKEN BAR
```
generalmimon, line 81: This looks good, by the way 👍

kaitai/kaitaistream.cpp
```diff
@@ -691,9 +693,143 @@ std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc)
...
+    }
+
+    switch (codepage) {
+    case KAITAI_CP_UTF16LE:
+        // If our source is already UTF16LE, just copy it
+        utf16_len = src_len / 2;
+        utf16 = std::wstring((wchar_t*)src.c_str(), utf16_len);
+        break;
+    case KAITAI_CP_UTF16BE:
+        // If our source is in UTF16BE, convert it to UTF16LE by swapping bytes
```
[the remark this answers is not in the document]
GreyCat: Fixed, thanks!

kaitai/kaitaistream.cpp
```diff
@@ -691,9 +693,143 @@ std::string kaitai::kstream::bytes_to_str(std::string src, std::string src_enc)
...
+        }
+
+        // Convert to UTF16 string
+        utf16 = std::wstring(utf16_len, L'\0');
+        if (MultiByteToWideChar(codepage, 0, src.c_str(), src_len, &utf16[0], utf16_len) == 0) {
+            throw std::runtime_error("bytes_to_str: MultiByteToWideChar conversion error");
+        }
+    }
+
+    // Step 3: convert UTF16 string to UTF8 string
```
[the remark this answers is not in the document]
GreyCat: Fixed, thanks!

kaitai/kaitaistream.h
```diff
@@ -319,6 +319,32 @@ class kstream {
 
     static void unsigned_to_decimal(uint64_t number, char *buffer);
 
+#ifdef KS_STR_ENCODING_WIN32API
+    enum {
+        KAITAI_CP_UNSUPPORTED = -1,
+        KAITAI_CP_UTF16LE = -2,
+        KAITAI_CP_UTF16BE = -3,
+    };
```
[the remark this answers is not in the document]
GreyCat: Thanks for looking this up! Let's keep it for now, until someone will prove that it's broken and there's indeed some C++98 compiler that needs absence of comma.

kaitai/kaitaistream.h
```diff
@@ -165,7 +165,7 @@ class kstream {
 
     static std::string bytes_strip_right(std::string src, char pad_byte);
     static std::string bytes_terminate(std::string src, char term, bool include);
-    static std::string bytes_to_str(std::string src, std::string src_enc);
+    static std::string bytes_to_str(const std::string src, const char *src_enc);
```
generalmimon, line 168: @GreyCat As evidenced by our CI dashboard (https://ci.kaitai.io/), this change breaks half of our tests for target `cpp_stl_11/msvc141_windows_x64` (see https://github.com/kaitai-io/ci_artifacts/commit/cf98b22205e3cb483fe2fdc11ec64a84f7913914#diff-ddeb848e50b284745134d0b3371b48f0cf63c52ed1d2f1134ff625c1217b57a3). The `ci.json` diff starts as:

```diff
diff --git 1/ci_json-cpp_stl_11-msvc141_windows_x64-f8948e3f-good.json 2/ci_json-cpp_stl_11-msvc141_windows_x64-cf98b222-bad.json
index d9b49a7..fe807ae 100644
--- 1/ci_json-cpp_stl_11-msvc141_windows_x64-f8948e3f-good.json
+++ 2/ci_json-cpp_stl_11-msvc141_windows_x64-cf98b222-bad.json
@@ -1,13 +1,13 @@
 {
   "$meta": {
     "lang": "cpp_stl",
-    "timestamp": "2023-07-08T22:54:28Z",
+    "timestamp": "2023-07-14T19:03:17Z",
     "ci": {
       "ci": "appveyor",
-      "build_id": "47500100",
-      "job_id": "a5x62yaygwv8fuxc",
+      "build_id": "47552134",
+      "job_id": "1rj2gtnxn51dxib8",
       "job_number": "3",
-      "url": "https://ci.appveyor.com/project/kaitai-io/ci-targets/builds/47500100/job/a5x62yaygwv8fuxc"
+      "url": "https://ci.appveyor.com/project/kaitai-io/ci-targets/builds/47552134/job/1rj2gtnxn51dxib8"
     }
   },
   "BcdUserTypeBe": "passed",
@@ -29,13 +29,13 @@
   "BitsUnalignedB64Le": "passed",
   "BufferedStruct": "passed",
   "BytesPadTerm": "passed",
-  "CastNested": "passed",
+  "CastNested": "format_build_failed",
   "CastToImported": "passed",
   "CastToTop": "passed",
   "CombineBool": "passed",
   "CombineBytes": "passed",
   "CombineEnum": "passed",
-  "CombineStr": "passed",
+  "CombineStr": "format_build_failed",
   "Debug0": "passed",
   "DebugArrayUser": "passed",
   "DebugEnumName": "passed",
@@ ... @@
```

The error is the following (https://ci.appveyor.com/project/kaitai-io/ci-targets/builds/47552134/job/1rj2gtnxn51dxib8?fullLog=true#L492):

```cpp
  cast_nested.cpp
C:\projects\ci-targets\tests\compiled\cpp_stl_11\cast_nested.cpp(90):
  error C2664: 'std::string kaitai::kstream::bytes_to_str(const std::string,const char *)':
    cannot convert argument 2 from 'std::string' to 'const char *'
    [C:\projects\ci-targets\tests\compiled\cpp_stl_11\bin\ks_tests.vcxproj]
  C:\projects\ci-targets\tests\compiled\cpp_stl_11\cast_nested.cpp(90):
    note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
```

Line [`compiled/cpp_stl_11/cast_nested.cpp:90`](https://github.com/kaitai-io/ci_targets/blob/bfe23a8fc97762d2472c018903886769ac04a69b/compiled/cpp_stl_11/cast_nested.cpp#L90) is:

```cpp
    m_value = kaitai::kstream::bytes_to_str(m__io->read_bytes_term(0, false, true, true), std::string("ASCII"));
```

But this apparently isn't a MSVC-specific issue. I've just checked [`cpp_stl_11/gcc4.8`](https://github.com/kaitai-io/ci_artifacts/tree/cpp_stl_11/gcc4.8) and the impact looks very much the same (see https://github.com/kaitai-io/ci_artifacts/commit/4ab4ae154437af18582d78696369799e604a134e#diff-ddeb848e50b284745134d0b3371b48f0cf63c52ed1d2f1134ff625c1217b57a3):

```diff
diff --git 1/ci_json-cpp_stl_11-gcc4.8-00689cd1-good.json 2/ci_json-cpp_stl_11-gcc4.8-4ab4ae15-bad.json
index fd80bd6..942318f 100644
--- 1/ci_json-cpp_stl_11-gcc4.8-00689cd1-good.json
+++ 2/ci_json-cpp_stl_11-gcc4.8-4ab4ae15-bad.json
@@ -1,13 +1,13 @@
 {
   "$meta": {
     "lang": "cpp_stl",
-    "timestamp": "2023-07-08T22:37:56Z",
+    "timestamp": "2023-07-14T18:46:24Z",
     "ci": {
       "ci": "github",
-      "run_id": "5496725152",
-      "run_number": "254",
-      "job_id": "14883558372",
-      "url": "https://github.com/kaitai-io/ci_targets/actions/runs/5496725152/jobs/10016910686"
+      "run_id": "5557284492",
+      "run_number": "255",
+      "job_id": "15054460788",
+      "url": "https://github.com/kaitai-io/ci_targets/actions/runs/5557284492/jobs/10150864944"
     }
   },
   "BcdUserTypeBe": "passed",
@@ -29,13 +29,13 @@
   "BitsUnalignedB64Le": "passed",
   "BufferedStruct": "passed",
   "BytesPadTerm": "passed",
-  "CastNested": "passed",
+  "CastNested": "format_build_failed",
   "CastToImported": "passed",
   "CastToTop": "passed",
   "CombineBool": "passed",
   "CombineBytes": "passed",
   "CombineEnum": "passed",
-  "CombineStr": "passed",
+  "CombineStr": "format_build_failed",
   "Debug0": "passed",
   "DebugArrayUser": "passed",
   "DebugEnumName": "passed",
@@ ... @@
```

The [`cpp_stl_11/gcc4.8`](https://github.com/kaitai-io/ci_artifacts/tree/cpp_stl_11/gcc4.8) target isn't visible at https://ci.kaitai.io/ though, because the CI dashboard is configured to display <code>cpp_stl_11/gcc4.8<ins><strong>_linux</strong></ins></code>, see [`src/App.vue:63`](https://github.com/kaitai-io/kaitai_ci_ui/blob/75212820e3322e6e8937ca161a200db01361b5c6/src/App.vue#L63):

```json
  ["cpp_stl_11", "gcc4.8_linux"],
```

PR #63, opened 2023-07-25
Use targets in CMake instead of variables

maksim-petukhov:
This PR replaces manual searching for Iconv with a find_package because there is [a find module for Iconv in CMake](https://cmake.org/cmake/help/latest/module/FindIconv.html) which is available since version 3.11. It is not merely a step towards modernization, it allows to cross-build this library when using Android NDK. This PR also sets 3.11 as minimum required version of CMake which is a rather reasonable because versions of CMake > 3.11 [are widely available in Linux distros](https://pkgs.org/download/cmake) (it goes without saying that you can also just download latest CMake version and install on any Linux distro). Zlib now is also linked as a target instead of using variables but this target [is available since CMake 3.1](https://cmake.org/cmake/help/latest/module/FindZLIB.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,20 +1,11 @@
-cmake_minimum_required (VERSION 3.3)
+cmake_minimum_required (VERSION 3.11)
 project (kaitai_struct_cpp_stl_runtime CXX)
 enable_testing()
 
 set (CMAKE_INCLUDE_CURRENT_DIR ON)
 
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks for this contribution, I believe this passes minimal bars that we had installed here.

By any chance you'll be interested in contributing CI flow that will build this (and prove that it works) in Android NDK context?

PR #64, opened 2023-07-27
Added ICU charset conversion implementation

GreyCat:


.github/workflows/build.yml
```diff
@@ -9,11 +9,22 @@ on:
...
+        encoding:
+          - ICONV
+          - ICU
+    env:
+      ENCODING_TYPE: ${{matrix.encoding}}
     steps:
       - uses: actions/checkout@v3
       - name: restore
         run: |
           sudo apt-get install -y libgtest-dev
```
generalmimon, line 23: I see I didn't catch this before, but we should always run `sudo apt-get update` before installing any packages. See for example the [Installing software on Ubuntu runners](https://docs.github.com/en/enterprise-cloud@latest/actions/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners) section in the official GitHub Actions docs:

> **Note**: Always run `sudo apt-get update` before installing a package. In case the apt index is stale, this command fetches and re-indexes any available packages, which helps prevent package installation failures.

GreyCat: Looks like fixed in e5e45dbb.

GreyCat:
@generalmimon If you want to take a closer look, this is now ready for review.

Should be relatively contained change. I plan to squash everything on merge.

generalmimon:
It would be good to check if we can apply something from [Recommended Build Options](https://unicode-org.github.io/icu/userguide/icu4c/build.html#recommended-build-options). The [How To Use ICU](https://unicode-org.github.io/icu/userguide/icu/howtouseicu.html#overview) section states:

> For C++, note that there are [Recommended Build Options](https://unicode-org.github.io/icu/userguide/icu4c/build.html#recommended-build-options) (both for normal use and for ICU as system-level libraries) which are not default simply for compatibility with older ICU-using code.

GreyCat:
> It would be good to check if we can apply something from [Recommended Build Options](https://unicode-org.github.io/icu/userguide/icu4c/build.html#recommended-build-options).

As far as I can tell, this speaks to setting up some defines to make use of namespaces. Namespaces per se will be only used in C++ APIs (which by themselves need to be activated by `U_SHOW_CPLUSPLUS_API` define), and we only use C APIs, which seem to be not affected by those.

Can you clarify what exactly do you want to see in build options for these?

kaitai/kaitaistream.cpp
```diff
@@ -1197,6 +1197,112 @@ std::string kaitai::kstream::bytes_to_str(const std::string src, int codepage) {
...
+        nullptr,
+        &err);
+    if (U_FAILURE(err)) {
+        ucnv_close(conv);
+        ucnv_close(utf8Conv);
+        throw illegal_seq_in_encoding(u_errorName(err));
+    }
+
+    // Allocate buffer for UTF-16 intermediate representation
+    const int32_t uniStrCapacity = UCNV_GET_MAX_BYTES_FOR_STRING(src.length(), ucnv_getMaxCharSize(conv));
```
generalmimon, line 1239: Is `ucnv_getMaxCharSize(conv)` guaranteed to be sufficient here? I'm not sure, because the documentation of [`ucnv_getMaxCharSize()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#af43eaa49c8f0e9bd5c63ace95b014d8b) says the following:

> Returns the maximum number of bytes that are output per UChar in conversion <ins>**from Unicode**</ins> using this converter.
> 
> The returned number can be used with UCNV_GET_MAX_BYTES_FOR_STRING to calculate the size of a target buffer for conversion <ins>**from Unicode**</ins>.

Whereas here we use it for a target buffer for conversion **to** Unicode (UTF-16), which is the opposite.

The _examples of returned values_ also illustrate the possible problem nicely:

> Examples for returned values:
> 
> - SBCS charsets: 1
> - Shift-JIS: 2
> - (...)

SBCS apparently stands for [single-byte character set](https://en.wikipedia.org/wiki/SBCS) (I've never heard this acronym before). This makes sense: if we take [ISO-8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Code_page_layout) as an example, then any `UChar` (UTF-16 character, also an unsigned 16-bit integer) valid in this encoding will be converted to  just a single byte. However, obviously no character in this charset maps to just 1 byte in UTF-16 (in fact, none does).

It seems that this mismatch in meaning wasn't caught by our tests because you're also using [`UCNV_GET_MAX_BYTES_FOR_STRING`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#aa3d7e4ae84f8a95b9735ed3491cdb77e) incorrectly (in a way that doesn't agree with the documentation):

> <dl>
> <dt>Parameters</dt>
> <dd><table>
> <tr><th align="left">length</th><td>Number of UChars to be converted.</td></tr>
> <tr><th align="left">maxCharSize</th><td>Return value from <a href="https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#af43eaa49c8f0e9bd5c63ace95b014d8b"><strong>ucnv_getMaxCharSize()</strong></a> for the converter that will be used.</td></tr>
> </table></dd></dl>

So the first parameter of the macro should be number of [`UChar`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/umachine_8h.html#a6bb9fad572d65b305324ef288165e2ac)s, i.e. UTF-16 _code units_ or unsigned 16-bit integers. However, you pass `src.length()`, which is the number of bytes.

CMakeLists.txt
```diff
@@ -37,6 +44,11 @@ if(Iconv_FOUND)
     target_link_libraries(${PROJECT_NAME} PRIVATE Iconv::Iconv)
 endif()
 
+if(ICU_FOUND)
+    target_include_directories(${PROJECT_NAME} PRIVATE ${ICU_INCLUDE_DIRS})
+    target_link_libraries(${PROJECT_NAME} PRIVATE ${ICU_LIBRARIES})
```
generalmimon, line 49: This is the old-school way of handling dependencies in CMake and I believe it is deprecated in favor of [targets](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#imported-targets). See also https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/63 and https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/.

generalmimon (changes requested):

kaitai/kaitaistream.cpp
generalmimon: According to the docs for [`ucnv_toUChars()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#ae1049fcb893783c860fe0f9d4da84939):

> The maximum output buffer capacity required (barring output from callbacks) will be 2\*srcLength (each char may be converted into a surrogate pair).

So we should probably use this instead.

generalmimon: Hmm, after thinking about it for a while, I've convinced myself that this calculated capacity should (coincidentally) always be enough, even it it's clearly incorrect and should definitely be fixed.

In theory, the only scenario in which it could underallocate memory would be for a SBCS (_single-byte character set_) with a character outside the BMP (Basic Multilingual Plane) - then `src.length()` would be `1` (assuming that only that one character is converted) and `ucnv_getMaxCharSize(conv)` would also be `1` (since it is a SBCS), so `uniStrCapacity` would be `1` and thus the `uniStr` array would only have space for 1 `UChar` (16-bit code unit). But since all Unicode characters **outside the BMP** (i.e. supplementary characters, U+10000 and greater) require 2 `UChars` to be represented in UTF-16 (see https://en.wikipedia.org/wiki/UTF-16#Code_points_from_U+010000_to_U+10FFFF), this would lead to an error.

However, such SBCS apparently doesn't exist, because BMP pretty much covers all the common characters that any sane (real-world) SBCS would use. The most popular supplementary characters (U+10000 and above) are emojis, but of course there is no real-world SBCS with emojis, since all SBCSs predate emojis by a lot.

So underallocation isn't a problem here, but overallocation by a certain factor is quite likely. For example, if we're decoding a UTF-8 string, then `ucnv_getMaxCharSize(conv)` returns 3 - see [`ucnv_getMaxCharSize()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#af43eaa49c8f0e9bd5c63ace95b014d8b) docs:

> Examples for returned values:
> 
> - (...)
> - UTF-8: 3 (3 per BMP, 4 per surrogate pair)

So `uniStr` will be allocated with a capacity of at least `3 * src.length()`, while `2 * src.length()` is already guaranteed be enough. As mentioned [above](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/64#discussion_r2033967746), each character of any encoding could be worst-case converted into a surrogate pair (a pair of two 16-bit code units, i.e. 2 `UChars`), but it never gets any worse.

kaitai/kaitaistream.cpp
```diff
@@ -1197,6 +1197,112 @@ std::string kaitai::kstream::bytes_to_str(const std::string src, int codepage) {
...
+
+    // Open UTF-8 converter
+    UConverter* utf8Conv = ucnv_open("UTF-8", &err);
+    if (U_FAILURE(err)) {
+        ucnv_close(conv);
+        throw bytes_to_str_error(u_errorName(err));
+    }
+
+    // Configure source converter to stop on illegal sequences
+    err = U_ZERO_ERROR;
```
generalmimon, line 1224: I don't think these `err = U_ZERO_ERROR` assignments are necessary, because we check the value of `err` after every call to an ICU function and throw an exception `if (U_FAILURE(err))`. Therefore, we know that if we get past any `if (U_FAILURE(err)) { ... }` block, we know that `err` indicates success.

Strictly speaking, it might not be equal to `U_ZERO_ERROR = 0`, because negative `err` values are [used for warnings](https://github.com/unicode-org/icu/blob/2d029329c82c7792b985024b2bdab5fc7278fbc8/icu4c/source/common/unicode/utypes.h#L421-L439) and `U_ZERO_ERROR` is only used when there is no warnings. But warnings don't represent errors, so they will not cause ICU functions to exit if they see one - only errors do, see [`UErrorCode` API docs](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/utypes_8h.html#a3343c1c8a8377277046774691c98d78c):

> Note: By convention, ICU functions that take a reference (C++) or a pointer (C) to a UErrorCode first test:
> 
> ```c
> if (U_FAILURE(errorCode)) { return immediately; }
> ```
> 
> so that in a chain of such functions the first one that sets an error code causes the following ones to not perform any operations.

Note that `U_FAILURE` is defined like this - [`icu4c/source/common/unicode/utypes.h:713-717`](https://github.com/unicode-org/icu/blob/2d029329c82c7792b985024b2bdab5fc7278fbc8/icu4c/source/common/unicode/utypes.h#L713-L717):

```c
    /**
     * Does the error code indicate a failure?
     * @stable ICU 2.0
     */
#   define U_FAILURE(x) ((x)>U_ZERO_ERROR)
```

---

Bottom line, I'd remove them because they are unnecessary.

kaitai/kaitaistream.cpp
```diff
@@ -1197,6 +1197,112 @@ std::string kaitai::kstream::bytes_to_str(const std::string src, int codepage) {
...
+        uniStr,
+        uniStrCapacity,
+        src.c_str(),
+        src.length(),
+        &err);
+    if (U_FAILURE(err)) {
+        delete[] uniStr;
+        ucnv_close(conv);
+        ucnv_close(utf8Conv);
+        throw illegal_seq_in_encoding(u_errorName(err));
```
generalmimon, line 1255: This is the only place where `illegal_seq_in_encoding` is thrown when running our test suite, and the only place where it actually makes sense to throw it. It makes no sense when handling potential errors from `ucnv_setToUCallBack`, `ucnv_setFromUCallBack` or `ucnv_fromUChars` - if these fail, it doesn't indicate an illegal sequence in the input string (instead, it would indicate a bug in our `bytes_to_str` implementation, as none of these should normally fail regardless of the user input).

Therefore, `throw illegal_seq_in_encoding` should only remain here and `throw bytes_to_str_error` should be used everywhere else.

CMakeLists.txt
generalmimon: I've just addressed this. It was pretty straightforward - I found this at https://cmake.org/cmake/help/v3.31/module/FindICU.html:

> Imported targets:
> 
> `ICU::<C>`
> 
> Where `<C>` is the name of an ICU component, for example `ICU::i18n`; `<C>` is lower-case.

Also, I noticed that the `io` component in `find_package(ICU COMPONENTS uc io)` was not used, so I removed it. We clearly don't use anything from this list:

```console
pp@DESKTOP-89OPGF3:/mnt/c/temp/kaitai_struct/runtime/cpp_stl$ nm -D --defined-only --demangle /usr/lib/x86_64-linux-gnu/libicuio.so
0000000000005460 T ufmt_64tou(char16_t*, int*, unsigned long, unsigned char, signed char, int)
00000000000058f0 T ufmt_uto64(char16_t const*, int*, signed char)
0000000000005440 T ufmt_isdigit(char16_t, int)
0000000000007a20 T ufile_flush_io(UFILE*)
00000000000053f0 T ufmt_digitvalue(char16_t)
00000000000079c0 T ufile_close_translit(UFILE*)
0000000000007880 T ufile_flush_translit(UFILE*)
0000000000007da0 T ufile_fill_uchar_buffer(UFILE*)
0000000000005ad0 T ufmt_defaultCPToUnicode(char const*, int, char16_t*, int)
00000000000057b0 T ufmt_ptou(char16_t*, int*, void*, signed char)
0000000000005970 T ufmt_utop(char16_t const*, int*)
000000000000a110 T icu_70::operator<<(std::ostream&, icu_70::UnicodeString const&)
000000000000a290 T icu_70::operator>>(std::istream&, icu_70::UnicodeString&)
00000000000050f0 T u_fadopt_70
0000000000007af0 T u_fclose_70
0000000000005220 T u_feof_70
0000000000007a50 T u_fflush_70
00000000000053c0 T u_fgetConverter_70
00000000000053d0 T u_fgetNumberFormat_70
0000000000008230 T u_fgetc_70
00000000000052d0 T u_fgetcodepage_70
0000000000008cf0 T u_fgetcx_70
0000000000005270 T u_fgetfile_70
0000000000005280 T u_fgetlocale_70
0000000000007fc0 T u_fgets_70
000000000000a060 T u_file_read_70
0000000000007b90 T u_file_write_70
00000000000074c0 T u_file_write_flush_70
00000000000050e0 T u_finit_70
0000000000005100 T u_fopen_70
000000000000a610 T u_fopen_u_70
0000000000007330 T u_fprintf_70
00000000000070d0 T u_fprintf_u_70
0000000000007d00 T u_fputc_70
0000000000007cc0 T u_fputs_70
0000000000007aa0 T u_frewind_70
0000000000008b90 T u_fscanf_70
00000000000089f0 T u_fscanf_u_70
0000000000005330 T u_fsetcodepage_70
0000000000005290 T u_fsetlocale_70
00000000000078b0 T u_fsettransliterator_70
0000000000005160 T u_fstropen_70
0000000000008d30 T u_fungetc_70
0000000000005e00 T u_get_stdout_70
0000000000004070 T u_locbund_close_70
0000000000004250 T u_locbund_getNumberFormat_70
0000000000003fd0 T u_locbund_init_70
00000000000073f0 T u_printf_70
0000000000005e70 T u_printf_parse_70
0000000000007190 T u_printf_u_70
00000000000082b0 T u_scanf_parse_70
0000000000006fa0 T u_snprintf_70
0000000000006d20 T u_snprintf_u_70
0000000000006ec0 T u_sprintf_70
0000000000006c60 T u_sprintf_u_70
0000000000008920 T u_sscanf_70
0000000000008780 T u_sscanf_u_70
0000000000007260 T u_vfprintf_70
0000000000007070 T u_vfprintf_u_70
0000000000008ab0 T u_vfscanf_70
00000000000089e0 T u_vfscanf_u_70
0000000000006dd0 T u_vsnprintf_70
0000000000006b70 T u_vsnprintf_u_70
0000000000006f80 T u_vsprintf_70
0000000000007050 T u_vsprintf_u_70
0000000000008840 T u_vsscanf_70
00000000000086a0 T u_vsscanf_u_70
00000000000041f0 T ucln_io_registerCleanup_70
0000000000008c50 T ufile_getch32_70
00000000000081c0 T ufile_getch_70
```

PR #65, opened 2023-09-28
Add support for building a static library of Kaitai Runtime

Pravila00:
# Changes

- Drop the SHARED keyword from add_library in CMakeLists.txt file.

# Context 

- Right now, the CMakeLists.txt generates always a shared library, without the possibility to users of generating a static library instead of a shared one.
- A shared library is built due to the SHARED keyword usage in add_library CMake function. 
- When no SHARED keyword is specified, the following behaviour happens according to CMake documentation at [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html): 
    -  `If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON` 
    - Given this piece of information, users will be able to build a static or a dynamic library through the variable [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS)


GreyCat:
@Pravila00 Windows CI builds seem to be failing now, as [build.ps1](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/master/.build/build.ps1#L34) relies on kaitai_struct_cpp_stl_runtime.dll being generated and now it looks like Windows default changes to static .lib generation.

I see several ways around this one:

* Change Windows CI script so it will set [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) and keep .dll copying.
* Remove .dll copying and stick with static libs on Windows from now on.
* Refactor GH Actions in a way that would allow testing both static+shared on Windows (probably a real overkill).

Pravila00:
Thanks for the feedback @GreyCat!

I have decided that we could continue building the dynamic library in the CI workflow.
For consistency purposes, I have activated the BUILD_SHARE_LIBS variable in Linux and Windows scripts.

GreyCat (approved):

generalmimon (commented):
For some reason, CMake's documentation suggests to add `BUILD_SHARED_LIBS` to the project using the [`option()`](https://cmake.org/cmake/help/latest/command/option.html#command:option) command, see:

1. https://cmake.org/cmake/help/latest/guide/tutorial/Selecting%20Static%20or%20Shared%20Libraries.html

   > In this section we will show how the [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) variable can be used to control the default behavior of [`add_library()`](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library), and allow control over how libraries without an explicit type (`STATIC`, `SHARED`, `MODULE` or `OBJECT`) are built.
   > 
   > To accomplish this we need to add [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) to the top-level `CMakeLists.txt`. <ins>**We use the [`option()`](https://cmake.org/cmake/help/latest/command/option.html#command:option) command as it allows users to optionally select if the value should be `ON` or `OFF`.**</ins>

2. https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

   > If present and true, this will cause all libraries to be built shared unless the library was explicitly added as a static library. <ins>**This variable is often added to projects as an [`option()`](https://cmake.org/cmake/help/latest/command/option.html#command:option) so that each user of a project can decide if they want to build the project using shared or static libraries.**</ins>

Do you know if there is any advantage of doing that?

Pravila00:
Hi @generalmimon ,

In reality, adding BUILD_SHARED_LIBS as option or not, doesn't have an impact, as it is a **global flag**.

Setting it as an option, or passing the value to the CMake command line function as `-DBUILD_SHARE_LIBS=ON/OFF` have the same output.

Nonetheless, as you mentioned it is often added as an option, so I am okay with setting it is an option.
Moreover, this will give visibility to users about the usage of BUILD_SHARE_LIBS, in case they were not aware of this.

I've just added a commit including it as an option, thanks a lot for your feedback!

generalmimon (approved):
Looks good to me, thanks!

PR #68, opened 2023-10-26
remove template id from constructor/destructor

toge:
In 0.10.1, I met compilation errors on gcc 13:
```
In file included from kaitai/kaitaistream.cpp:2:
kaitai/exceptions.h:118:35: error: expected unqualified-id before ‘const’
  118 |     validation_not_equal_error<T>(const T& expected, const T& actual, kstream* io, const std::string src_path):
      |                                   ^~~~~
kaitai/exceptions.h:118:35: error: expected ‘)’ before ‘const’
  118 |     validation_not_equal_error<T>(const T& expected, const T& actual, kstream* io, const std::string src_path):
      |                                  ~^~~~~
      |                                   )
kaitai/exceptions.h:127:13: error: template-id not allowed for destructor
  127 |     virtual ~validation_not_equal_error<T>() KS_NOEXCEPT {};
      |             ^
...
```

They are caused by `<T>` in constructor/destructor.
I want to remove it.

I confirmed gcc/clang/MSVC work fine after deleteing `<T>`.

### DIFF
--- a/kaitai/exceptions.h
+++ b/kaitai/exceptions.h
@@ -115,7 +115,7 @@ protected:
 template<typename T>
 class validation_not_equal_error: public validation_failed_error {
 public:
-    validation_not_equal_error<T>(const T& expected, const T& actual, kstream* io, const std::string src_path):
+    validation_not_equal_error(const T& expected, const T& actual, kstream* io, const std::string src_path):
         validation_failed_error("not equal", io, src_path),
         m_expected(expected),
         m_actual(actual)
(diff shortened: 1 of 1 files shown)


generalmimon:
Reproduced on gcc 11.4, gcc 12.3 and gcc 13.2, but **only** with the `-std=c++20` flag. Not reproduced on gcc 10.5 or clang 17.0.1.

generalmimon (approved):
@toge Thanks!

PR #74, opened 2024-05-27
Support FreeBSD 14.0 + NetBSD 10.0 + add testing

GreyCat:
Adds proven support for FreeBSD and NetBSD. Uses GH Actions for 2 new workflows (which internally download an image for a VM and runs it in qemu emulator on Ubuntu host).

### DIFF
--- a/tests/unittest.cpp
+++ b/tests/unittest.cpp
@@ -4,7 +4,10 @@
 #include <gtest/gtest.h>
 #endif
 
+// Necessary for proper detection of *BSD
+#if !defined(_MSC_VER)
 #include <sys/param.h>
+#endif
 
 #include "kaitai/kaitaistream.h"
 #include "kaitai/exceptions.h"
--- a/tests/unittest.cpp
+++ b/tests/unittest.cpp
@@ -4,6 +4,8 @@
 #include <gtest/gtest.h>
 #endif
 
+#include <sys/param.h>
+
 #include "kaitai/kaitaistream.h"
 #include "kaitai/exceptions.h"
 #include <sstream>
@@ -359,7 +361,6 @@ TEST(KaitaiStreamTest, bytes_to_str_invalid_seq_euc_jp_too_short)
     }
 }
 
-
 TEST(KaitaiStreamTest, bytes_to_str_invalid_seq_gb2312_too_short)
 {
     try {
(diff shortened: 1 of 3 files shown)


GreyCat:
Hey @saper @generalmimon @reezer,

I went ahead and adjusted defines in a way that it makes it work on modern FreeBSD + modern NetBSD.

This fixes problems with compilation + problems with defines resulting in floating point reading functions misbehaving.

The only caveat remaining is iconv: it looks like starting circa FreeBSD 10.0 and sometimes in recent NetBSD, new iconv-like engine was adopted, something called [CITRUS](http://citrus.bsdclub.org/index-en.html). The problem with it, it seems to be supporting much smaller set of encodings, and, it's quite forgiving, which makes our tests like "convert B0 30 in GB2312 to UTF-8" not report an error, while it clearly should.

My only idea is to specifically detect CITRUS and disable the relevant test. Any objections?

saper:
Excellent find. I think that's a pretty good idea to disable the test. Great work, thank you very much!

saper:
Bruno Haible @bhaible identified and reported multiple issues related to the Citrus iconv being too permissive:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272382
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272384
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272386

We should definitely skip that test for now.

bhaible:
> Bruno Haible @bhaible identified and reported multiple issues related to the Citrus iconv being too permissive:
> 
> https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272382
> https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272384
> https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272386

"too permissive" is an adequate description for the third among these bugs. For the other two, "broken" is a better summary.

reezer:
I agree and looks good to me as well. Thanks for picking this up!

tests/unittest.cpp
```diff
@@ -4,6 +4,11 @@
 #include <gtest/gtest.h>
 #endif
 
+// Necessary for proper detection of *BSD
+#if !defined(_MSC_VER)
+#include <sys/param.h>
+#endif
```
generalmimon, line 10: @GreyCat I suspect this isn't actually necessary since you're only using the `__FreeBSD__` and `__NetBSD__` macros, not the `BSD` macro which indeed requires `#include <sys/param.h>`: https://sourceforge.net/p/predef/wiki/OperatingSystems/#bsd-environment

I think this unnecessary include should be removed. For now, I guess this condition is fine, because the test exclusions should be minimal and these are the only BSD variants that we know are problematic:

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/7b4c29a16b63d3c7738c124423a0087e97ad7834/tests/unittest.cpp#L383

GreyCat: I recall in my testing that it actually required that include, but maybe I've indeed got something crossed, so if you have an idea what we can eliminate/optimize here — absolutely feel free to go ahead and change :)

generalmimon: I think (at least I haven't seen any single reference on the internet that would suggest otherwise) that `sys/param.h` only defines `BSD` (see [FreeBSD's `sys/param.h`](https://github.com/freebsd/freebsd-src/blob/35e941ce5520b8ef710bb388bd63460aa0ce63eb/sys/sys/param.h#L42), [NetBSD's `sys/param.h`](https://github.com/NetBSD/src/blob/5680363e133011c306e0a21efff4e74c17f9fd4c/sys/sys/param.h#L49)), whereas `__FreeBSD__` and `__NetBSD__` are predefined macros by the compiler, so you don't need to include anything to use them.

PR #75, opened 2025-01-24
Make KS_ZLIB flag public

ithron:
When using the runtime with zlib support with cmake, the required KS_ZLIB flags is not propagated to the linking target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ target_include_directories(${PROJECT_NAME} INTERFACE ${PROJECT_SOURCE_DIR})
 
 if (ZLIB_FOUND)
     target_link_libraries(${PROJECT_NAME} PRIVATE ZLIB::ZLIB)
-    target_compile_definitions(${PROJECT_NAME} PRIVATE -DKS_ZLIB)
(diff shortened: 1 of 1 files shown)


generalmimon:
@ithron Makes sense, thanks for the catch. Could you please fix the typo in the commit message? (It should be `KS_ZLIB`, not `KS_ZLIP`.)

I'm thinking how we can test this in CI - this is a very easy mistake to make, so we should have some sort of safety net to ensure it doesn't happen again.

If I understand correctly, the problem with `-DKS_ZLIB` having `PRIVATE` scope is that if you include `kaitai/kaitaistream.h` from an external project, then `KS_ZLIB` will not be defined and thus the `process_zlib` method will not exist?

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/33502322683ab40353c7c956d382fc2b09e6af70/kaitai/kaitaistream.h#L216-L224

ithron:
Fixed the typo.

Yes, that's exactly the issue.
I think the issue went unnoticed because the unit tests don't use the zlib function. Since the unit tests are linked against the runtime using cmake, they would not compile if one of the tests tried to access process_zlib().

generalmimon (approved):
@ithron Thanks!

PR #76, opened 2025-03-02
to_string(): replace template with 6 overloads, simplify

generalmimon:
As mentioned in the code comment in `kaitai/kaitaistream.cpp`, this approach follows what `std::to_string` is doing:
https://en.cppreference.com/w/cpp/string/basic_string/to_string

Avoiding templates and separating signed and unsigned cases has been already suggested by @webbnh in https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/pull/50.

The motivation for this change is that this new version of `to_string` also accepts values of unscoped enums that can be implicitly converted to integers, just like `std::to_string` does. For example, you can do `kaitai::kstream::to_string(FRUIT_APPLE)` just like you can do `std::to_string(FRUIT_APPLE)`, where `FRUIT_APPLE` is an unscoped enum member:

```cpp
enum fruit_t {
    FRUIT_APPLE = 2,
    FRUIT_ORANGE = 5
};
```

This simplifies the translation of expressions like `fruit::apple.to_i.to_s`, because it means that the `enum.to_i` operation can remain a no-op in the KSC:

[`CppTranslator.scala:193-194`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/12dbc32226eb9e94b76f803f3c8ff5f8943e5f8d/shared/src/main/scala/io/kaitai/struct/translators/CppTranslator.scala#L193-L194)

```scala
  override def enumToInt(v: expr, et: EnumType): String =
    translate(v)
```

Therefore, this change fixes the [EnumToI](https://github.com/kaitai-io/kaitai_struct_tests/blob/af0359aeb87233640fcd85b57850c712b6f03f63/formats/enum_to_i.ksy) and EnumToIInvalid tests for C++11, which currently fail with this error (https://github.com/kaitai-io/ci_artifacts/blob/786a45a62978409fd90fab124b5b1ad1f225a81e/test_out/cpp_stl_11/build-1.log#L1495-L1498):

```
/tests/compiled/cpp_stl_11/enum_to_i.cpp: In member function 'std::string enum_to_i_t::pet_1_i_to_s()':
/tests/compiled/cpp_stl_11/enum_to_i.cpp:65:48: error: no matching function for call to 'kaitai::kstream::to_string(enum_to_i_t::animal_t)'
   65 |     m_pet_1_i_to_s = kaitai::kstream::to_string(pet_1());
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
```

Eventually, we might switch to scoped enums (see https://github.com/kaitai-io/kaitai_struct/issues/959), which are not implicitly convertible to integers, so then it won't be possible to keep the `.to_i` operation implemented as a no-op, but for now (as long as we're using unscoped enums) it works.

### DIFF
--- a/.build/gtest-v1.8.1-fix-Werror-maybe-uninitialized.patch
+++ b/.build/gtest-v1.8.1-fix-Werror-maybe-uninitialized.patch
@@ -1,0 +1,21 @@
+diff --git a/googletest/src/gtest-death-test.cc b/googletest/src/gtest-death-test.cc
+index 0908355..d6f492e 100644
+--- a/googletest/src/gtest-death-test.cc
++++ b/googletest/src/gtest-death-test.cc
+@@ -1212,14 +1212,14 @@ static int ExecDeathTestChildMain(void* child_arg) {
+ static void StackLowerThanAddress(const void* ptr,
+                                   bool* result) GTEST_NO_INLINE_;
+ static void StackLowerThanAddress(const void* ptr, bool* result) {
+-  int dummy;
++  int dummy = 0;
+   *result = (&dummy < ptr);
+ }
+
+ // Make sure AddressSanitizer does not tamper with the stack here.
+ GTEST_ATTRIBUTE_NO_SANITIZE_ADDRESS_
+ static bool StackGrowsDown() {
+-  int dummy;
++  int dummy = 0;
+   bool result;
+   StackLowerThanAddress(&dummy, &result);
+   return result;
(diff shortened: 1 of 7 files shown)


webbnh:
Hi Guys!  Cool to see that you are still working on this stuff.  :-)

This change looks pretty nice!

generalmimon:
It seems by now that I've gotten all the reviews I could get (thanks @webbnh), so I'm merging this in :)

PR #92, opened 2026-06-03
GHA: bump to macOS 26 (was 15)

generalmimon:
BTW, `macos-latest` is currently still an alias of `macos-15`, but it will soon become `macos-26`: https://github.com/actions/runner-images/issues/14167

This means that `macos-26` should be stable enough for us to switch to.

### DIFF
--- a/.build/gtest-v1.17.0-fix-Werror-character-conversion.patch
+++ b/.build/gtest-v1.17.0-fix-Werror-character-conversion.patch
@@ -1,0 +1,22 @@
+diff --git a/googletest/include/gtest/gtest-printers.h b/googletest/include/gtest/gtest-printers.h
+index 198a7693..bbb0fa7e 100644
+--- a/googletest/include/gtest/gtest-printers.h
++++ b/googletest/include/gtest/gtest-printers.h
+@@ -521,11 +521,15 @@ GTEST_API_ void PrintTo(wchar_t wc, ::std::ostream* os);
+ 
+ GTEST_API_ void PrintTo(char32_t c, ::std::ostream* os);
+ inline void PrintTo(char16_t c, ::std::ostream* os) {
+-  PrintTo(ImplicitCast_<char32_t>(c), os);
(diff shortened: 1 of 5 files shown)


generalmimon:
Unfortunately, we got a build error in the `macos (20)` CI job:

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/actions/runs/29121972211/job/86458913076?pr=92#step:4:50

```
/usr/local/include/gtest/gtest-printers.h:528:35: error: implicit conversion from 'char8_t' to 'char32_t' may change the meaning of the represented code unit [-Werror,-Wcharacter-conversion]
  528 |   PrintTo(ImplicitCast_<char32_t>(c), os);
      |           ~~~~~~~~~~~~~           ^
1 error generated.
```

We currently install GoogleTest 1.16.0 if the C++20 standard is requested:

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/5a78e26c2aa662ffc74b9ae5e2704078157e6f9c/.build/install-gtest#L42-L46

v1.16.0 is no longer the latest version - https://github.com/google/googletest/releases/tag/v1.17.0 was released on Apr 30, 2025. However, it seems that simply updating to v1.17.0 would not fix the error, because it has been reported on GoogleTest 1.17.0 too: https://github.com/google/googletest/issues/4948

According to https://github.com/google/googletest/issues/4948#issuecomment-4909516485, it has been fixed in commit https://github.com/google/googletest/commit/fa8438ae6b70c57010177de47a9f13d7041a6328 on the `main` branch, but no version with this fix has been released yet.

Since we install GoogleTest from source, we can apply this fix as a patch, like we already patch GoogleTest 1.8.1 (the last release supporting C++98):

https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/c728ee444310f3864136745f7d426d8740bf9d10/.build/install-gtest#L67-L71

PR #96, opened 2026-06-11
GHA: run FreeBSD VM using cross-platform-actions/action

generalmimon:
Since https://github.com/cross-platform-actions/action/releases/tag/v1.1.0, `cross-platform-actions/action` allows (and encourages) using multiple steps, making CI run logs look more consistent with those of other jobs.

In addition, dropping the https://github.com/vmactions/freebsd-vm action means one less dependency we have to worry about.

To me, https://github.com/cross-platform-actions/action also seems to be better maintained and the code is better organized. In our case, it's also consistently faster (~1 min 20 seconds vs 2+ minutes with `vmactions/freebsd-vm`).

Closes #93

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -145,36 +145,36 @@ jobs:
         run: .build\run-unittest.ps1
   freebsd:
     runs-on: ubuntu-latest
+    defaults:
+      run:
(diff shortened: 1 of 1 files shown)


GreyCat (approved):
Looks great, thanks for looking into this!

